PR #126, opened 2024-01-04
Use CMS communication manager interface

SvizelPritula:
Currently, communication tasks in CMS mode use a custom interface to communicate with the manager.
This PR changes brings the interface in sync with CMS.

### DIFF
--- a/pisek/jobs/parts/solution.py
+++ b/pisek/jobs/parts/solution.py
@@ -33,7 +33,7 @@ from pisek.jobs.parts.solution_result import (
     SUBTASK_SPEC,
     solution_res_true,
 )
-from pisek.jobs.parts.judge import judge_job, RunJudge, RunBatchJudge
+from pisek.jobs.parts.judge import judge_job, RunJudge, RunCMSJudge, RunBatchJudge
 
 
 class SolutionManager(TaskJobManager):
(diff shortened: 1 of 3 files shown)


5K1PY:
Abstracted loading points to ``RunCMSJudge`` class so we don't keep two implementations of same thing.

pisek/jobs/parts/solution.py
```diff
@@ -401,79 +403,47 @@ def __init__(
...
+            os.mkfifo(fifo_to_solution)
+
+            # Open fifos to prevent blocking on future opens
+            os.open(fifo_from_solution, os.O_RDWR)
+            os.open(fifo_to_solution, os.O_RDWR)
+
+            fd_from_solution = os.open(fifo_from_solution, os.O_WRONLY)
+            fd_to_solution = os.open(fifo_to_solution, os.O_RDONLY)
+
+            points_file = self._output(self.input_name.replace(".in", ".judge"))
```
5K1PY, line 420: I don't particulary like this as this puts points files in outputs folder, however that's for sepearate issue.

5K1PY: See #127 for reference.

5K1PY (approved):

PR #131, opened 2024-01-20
Match behaviour of stubs to CMS

SvizelPritula:
Currently, `pisek` expects a stub or grader to be a single header file.
This does not match CMS, where a stub or grader is a separate compilation unit.

This change also allows one to specify any number of header available to both compilation units.


5K1PY:
This does break task using the old stub format, right?
We could at least edit ``pisek update`` so at least updating v1 tasks works.

SvizelPritula:
That's true, I'll fix the update tool. I don't know of any old tasks that use a stub with a version 2 config.

SvizelPritula:
I've updated the update tool.

5K1PY (approved):

PR #132, opened 2024-02-11
Redo paths

5K1PY:
Rewrote how we represent paths. Now instead of path string we have ``TaskPath`` instance that contains:
- Full path (for file operations)
- Path relative to task directory (for failures)
- Basename (for job names)

This way we can always choose which form we want to show instead of constant transformations

### DIFF
--- a/pisek/jobs/parts/program.py
+++ b/pisek/jobs/parts/program.py
@@ -204,7 +204,7 @@ class ProgramsJob(TaskJob):
         executable = TaskPath.executable_file(self._env, program.relpath)
         if not self._file_exists(executable):
             raise PipelineItemFailure(
-                f"Program {executable.relpath} does not exist, "
+                f"Program {executable:p} does not exist, "
                 f"although it should have been compiled already."
             )
         return executable
--- a/pisek/jobs/parts/program.py
+++ b/pisek/jobs/parts/program.py
@@ -70,28 +70,30 @@ class RunResult:
         res = tab(TaskHelper._short_text(text, max_lines, max_chars))
         return colored(res, env, "yellow")
 
-    def raw_stdout(self):
-        if isinstance(self.stdout_file, str):
-            return open(self.stdout_file).read()
+    def raw_stdout(self, access_file: Callable[[TaskPath], None]):
+        if isinstance(self.stdout_file, TaskPath):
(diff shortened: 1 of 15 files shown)


pisek/jobs/parts/program.py
```diff
@@ -193,36 +199,36 @@ def __init__(self, env: Env, name: str, **kwargs) -> None:
...
 
-    def _load_compiled(self, program: str) -> str:
+    def _load_compiled(self, program: TaskPath) -> TaskPath:
         """Loads name of compiled program."""
-        executable = self._executable(os.path.basename(program))
+        executable = TaskPath.executable_file(self._env, program.relpath)
         if not self._file_exists(executable):
             raise PipelineItemFailure(
-                f"Program {executable} does not exist, "
+                f"Program {executable.relpath} does not exist, "
```
SvizelPritula, line 207: Shouldn't this use the custom format specs?

pisek/jobs/parts/task_job.py
```diff
@@ -140,15 +67,29 @@ def _get_limits(self, program_type: ProgramType) -> dict[str, Any]:
...
+    ) -> list[TaskPath]:
+        files: list[str] = list(
+            sorted(
+                set(
+                    sum(
+                        (glob.glob(g, root_dir=directory.fullpath) for g in globs),
+                        start=[],
+                    )
+                )
+            )
```
SvizelPritula, line 88: This would probably be more readable if split into multiple statements.

pisek/paths.py
```diff
@@ -0,0 +1,171 @@
...
+INVALID_OUTPUTS_SUBDIR = "invalid/"
+OUTPUTS_SUBDIR = "outputs/"
+SANITIZED_SUBDIR = "sanitized/"
+LOG_SUBDIR = "log/"
+
+
+class TaskPath:
+    """Class representing a path to task file."""
+
+    def __init__(self, task_path, *path: str):
```
SvizelPritula, line 35: The type of `task_path` should be annotated.

SvizelPritula (changes requested):
Looks great, but there are a couple of small flaws.

SvizelPritula (approved):
LGTM

Issue #135, opened 2024-02-26
Hierarchical versions of config


5K1PY:
Generalize ``task_type`` to allow specifying config to get defaults from.

Relevant folder to find configs there would be found:
1. In enviroment variable if given
2. Finding upwards from current task directory (as git finds ``.git``)

Issue #134, opened 2024-02-26
Override limits per solution


5K1PY:
Let config user override limits in ``[solution_*]`` section.

PR #136, opened 2024-02-27
Pydantic env

5K1PY:
This is a big one.

- Rewrote env to pydantic to offer better type checking
- Implemented hierarchical configs (now only reading from current and defaults)
- Added new ! special values and @ autoexpand values to config

Lays the groundwork for #134 and #135.


pisek/server.py
```diff
@@ -120,9 +119,7 @@ def judge(
         return judging_res.points, judging_res.judge_stderr
 
     def _needs_generating(self, put: Optional[str]):
-        return put is None or not os.path.exists(
-            os.path.join(self.path, DATA_SUBDIR, put)
-        )
+        return put is None or not os.path.exists(os.path.join(self.path, "data/", put))
```
SvizelPritula, line 122: Shouldn't this use `TaskPath`?

pisek/utils/text.py
```diff
@@ -37,27 +30,23 @@ def pad_left(text: str, length: int, pad_char: str = " "):
...
 
 
-@no_color_variant
-def colored(msg: str, color: str) -> str:
-    # Recolors all white text to given color
-    col = getattr(Fore, color.upper())
-    msg = msg.replace(f"{Fore.RESET}", f"{Fore.RESET}{col}")
-    return f"{col}{msg}{Fore.RESET}"
+def warn(msg: str, err: type, strict: bool = False, no_colors: bool = False) -> None:
+    """Warn strict is False, otherwise raise error."""
```
SvizelPritula, line 49: This sentence makes no sense.

example-config
```diff
@@ -5,6 +5,10 @@
 # Move 'config' into a task folder 
 
 
+# Values begining with ! or @ are reserved and have special meaning:
+# - Only ! value is !unset, which is always invalid
```
SvizelPritula, line 9: What is it for?

pisek/env/task_config.py
```diff
@@ -0,0 +1,529 @@
...
+    contest_type: str
+    task_type: TaskType
+    fail_mode: FailMode
+
+    solutions_subdir: str
+    static_subdir: str
+    data_subdir: str
+
+    in_gen: str
+    checker: str
```
SvizelPritula, line 102: This should also be `Optional[str]`.

SvizelPritula (changes requested):
Looks great, but there are a couple of issues.

pisek/env/task_config.py
```diff
@@ -0,0 +1,529 @@
...
+                args[key] = default
+
+        section_names = configs.sections()
+
+        # Load subtasks
+        args["subtasks"] = subtasks = {}
+        for section_name in sorted(section_names):
+            if m := re.match(r"test(\d{2})", section_name):
+                num = int(m[1])
+                subtasks[num] = SubtaskConfig.load_dict(num, configs)
```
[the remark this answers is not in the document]
5K1PY: We already do that on lines 219-225.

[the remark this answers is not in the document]
SvizelPritula: Ok, great!

example-config
5K1PY: This is for #135. It does not make sense yet, but you should be able to force task author to set a key.

pisek/env/config_hierarchy.py
```diff
@@ -0,0 +1,104 @@
...
+
+class TaskConfigError(Exception):
+    pass
+
+
+class ConfigHierarchy:
+    def __init__(self, task_path: str) -> None:
+        self._config_path = os.path.join(task_path, CONFIG_FILENAME)
+        config_paths = [DEFAULTS_CONFIG, self._config_path]
+        self._configs = [ConfigParser() for _ in config_paths]
```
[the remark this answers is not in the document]
5K1PY: Would it? I don't see it.

example-config
SvizelPritula: As currently implemented, using !unset has the same effect on value resolution as simply not specifing the key, except that is skips checking other candidates. It will still search lower priority configs.

pisek/env/config_hierarchy.py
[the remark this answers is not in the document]
SvizelPritula: I would have written it like this:
```py
self._configs = []
for config, path in zip(self._configs, config_paths):
    config = ConfigParser()
    if not config.read(path):
        raise TaskConfigError(f"Missing config {path}. Is this task folder?")
    self._configs.append(config)
```
This avoids doing a `zip`. Which version looks better is up to debate.

example-config
5K1PY: I need to break twice, sorry..

pisek/server.py
5K1PY: I would leave that up to #137

SvizelPritula: Ok, fair.

pisek/env/base_env.py
```diff
@@ -0,0 +1,136 @@
...
+        self._logging = True
+
+        return accessed
+
+    def get_compound(self, key: str) -> Any:
+        """Get attribute that may be nested deeper and indexed."""
+        obj = self
+        for key_part in key.split("."):
+            if "[" in key_part:
+                var, index = key_part[:-1].split("[")
```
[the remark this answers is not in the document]
SvizelPritula: I still think it would be a good idea to add `assert key_part[-1] == ']'`.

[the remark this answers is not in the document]
5K1PY: Fair

SvizelPritula (approved):
LGTM

PR #139, opened 2024-02-29
Simplified TaskPaths and get rid of task_dir

### LINKED ISSUE #138: Rework TaskPaths and get rid of env.task_dir
Let's get rid of ``env.task_dir`` and let's execute all for current working directory. Additionally rework taskpaths to only contain path and name.

This would allow us cache runs independently from path we are testing them from. Also it makes dealing with ``TaskPath``s much less a hassle.

5K1PY:
Fixes #138.

### DIFF
--- a/pisek/env/env.py
+++ b/pisek/env/env.py
@@ -33,7 +33,6 @@ class Env(BaseEnv):
     Collection of environment variables for task testing
 
     Attributes:
-        task_dir: Directory of the task being tested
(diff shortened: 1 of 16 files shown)


SvizelPritula (approved):
Looks good to me.

`TaskPath` doesn't offer a lot of functionality anymore, but I like how it still makes type annotations more expressive.

PR #143, opened 2024-03-06
Visualize

### LINKED ISSUE #116: Rewrite visualization
Currently visualization is kind of a mess. There are bugs, it doesn't know parital results, can visualize only by time and so on.
A complete would be desirable.

Also generating not only text but graphs would be nice.

5K1PY:
Fixes #116

### DIFF
--- a/pisek/env/task_config.py
+++ b/pisek/env/task_config.py
@@ -36,7 +36,7 @@ from pisek.utils.text import eprint, colored, warn
 from pisek.env.base_env import BaseEnv
 from pisek.env.config_hierarchy import TaskConfigError, ConfigHierarchy
 from pisek.env.context import init_context
-from pisek.jobs.parts.solution_result import Verdict, SUBTASK_SPEC, verdict_true
+from pisek.jobs.parts.solution_result import SUBTASK_SPEC
 
 
 MaybeInt = Annotated[
@@ -273,31 +273,6 @@ class TaskConfig(BaseEnv):
 
         for i in range(self.subtasks_count):
             compute_subtask(i)
-
-    def evaluate_verdicts(
-        self, verdicts: list[Verdict], expected: str
-    ) -> tuple[bool, bool, Optional[int]]:
-        mode_quantifier = all if self.fail_mode == FailMode.all else any
-
-        result = True
-        definitive = True
-        breaker = None
-        quantifiers = [all, mode_quantifier]
-        for i, quant in enumerate(quantifiers):
-            oks = list(map(SUBTASK_SPEC[expected][i], verdicts))
-            ok = quant(oks)
-
-            result &= ok
-            definitive &= (
-                (quant == any and ok)
-                or (quant == all and not ok)
-                or (SUBTASK_SPEC[expected][i] == verdict_true)
-            )
-            if quant == all and ok == False:
-                breaker = oks.index(False)
-                break
-
-        return result, definitive, breaker
 
 
 class SubtaskConfig(BaseEnv):
--- a/pisek/env/task_config.py
+++ b/pisek/env/task_config.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from enum import StrEnum, auto
+import fnmatch
 from functools import cached_property
 from pydantic_core import PydanticCustomError, ErrorDetails
 from pydantic import (
@@ -283,6 +284,12 @@ class SubtaskConfig(BaseEnv):
     in_globs: ListStr
     all_globs: list[str] = []
     predecessors: list[int]
+
+    def in_subtask(self, filename: str) -> bool:
+        return any(fnmatch.fnmatch(filename, g) for g in self.all_globs)
+
+    def new_in_subtask(self, filename: str) -> bool:
+        return any(fnmatch.fnmatch(filename, g) for g in self.in_globs)
 
     @staticmethod
     def load_dict(number: int, configs: ConfigHierarchy) -> dict[str, Any]:
--- a/pisek/env/task_config.py
+++ b/pisek/env/task_config.py
@@ -341,6 +341,13 @@ class SubtaskConfig(BaseEnv):
 
         return list(sorted(set(predecessors)))
 
+    @model_validator(mode="after")
+    def validate_model(self):
+        if self.name == "@auto":
+            self.name = f"Subtask {self.num}"
+
+        return self
+
 
 class SolutionConfig(BaseEnv):
     """Configuration of one solution."""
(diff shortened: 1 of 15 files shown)


pisek/visualize.py
```diff
@@ -13,304 +9,330 @@
...
+        if new_verdict == Verdict.timeout:
+            new_verdict = Verdict.ok
 
+        return LoggedResult(
+            new_verdict, 1.0, result.time, result.test, result.original_verdict
+        )
+    if result.time > limit:
+        return LoggedResult(
+            Verdict.timeout, 0.0, result.time, result.test, result.original_verdict
+        )
```
SvizelPritula, line 106: This method will always set `points` to either `0.0` or `1.0`. Is this desired?

pisek/visualize.py
```diff
@@ -13,304 +9,330 @@
...
 
+        return SolutionResults(name, config, results)
+
+    def get_all(self) -> list[LoggedResult]:
+        return self._results
+
+    def get_by_subtask(self) -> dict[int, list[LoggedResult]]:
+        by_subtask: dict[int, list[LoggedResult]] = {
+            num: [] for num in self._config.subtasks
+        }
```
SvizelPritula, line 169: Couldn't this be a list?

pisek/visualize.py
```diff
@@ -13,304 +9,330 @@
...
-    return subtasks
-
+from typing import Optional
+
+from pisek.utils.text import pad, tab, colored
+from pisek.utils.terminal import terminal_width
+from pisek.jobs.parts.solution_result import Verdict
+from pisek.env.task_config import load_config, TaskConfig
+
+test_name_max_length = 0
```
SvizelPritula, line 23: Is a global variable really necessary?
Couldn't you calculate the maximum length in `visualize` and pass it as a parameter to `to_str`?

pisek/visualize.py
```diff
@@ -13,304 +9,330 @@
...
+                color = "red"
+            elif abs(percentage - 1) < 0.5:
+                color = "yellow"
+            else:
+                color = "green"
+
+            full_bar = colored("━", color)
+
+            return (
+                f"[{full_bar*bounded}{'━'*(segments-bounded)}|"
```
SvizelPritula, line 83: Idea: The non-colored bars could be a different character, say '┅' instead of '━'.

SvizelPritula (changes requested):
Looks great! It will be nice to have `visualize` working again.
There are a couple of small issues, though.

pisek/visualize.py
5K1PY: Not really. If neither case is true, we return original result. We could keep the original points, but this isn't consistent with new results.

We could not touch the points if you think that is better.

pisek/visualize.py
5K1PY: No. Same item can appear in different subtasks.

pisek/env/task_config.py
```diff
@@ -273,6 +274,31 @@ def compute_subtask(num: int) -> list[str]:
...
+        quantifiers = [all, mode_quantifier]
+        for i, quant in enumerate(quantifiers):
+            oks = list(map(SUBTASK_SPEC[expected][i], verdicts))
+            ok = quant(oks)
+
+            result &= ok
+            definitive &= (
+                (quant == any and ok)
+                or (quant == all and not ok)
+                or (SUBTASK_SPEC[expected][i] == verdict_true)
```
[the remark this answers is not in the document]
5K1PY: Imagine we are checking that subtask is `W` and it already answered wrong on one input. (``fail_mode=any``) Therefore the ``any`` part of ``SUBTASK_SPEC`` is definitive. However the ``all`` part will not be definitive and we will evaluate all inputs, even though solution already answered wrong.

(If this is too confusing, delete it and run it on ``fixtures/sum_cms``, it behaves as with ``--all-inputs``)

pisek/env/task_config.py
```diff
@@ -273,6 +274,31 @@ def compute_subtask(num: int) -> list[str]:
...
+            )
+            if quant == all and ok == False:
+                breaker = oks.index(False)
```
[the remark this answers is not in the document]
5K1PY: I don't see how it simplifies things.

pisek/visualize.py
5K1PY: I would keep `'┅'`'s special meaning to cutted bar.

pisek/visualize.py
SvizelPritula: Sorry, I missed the final `return`. This way makes sense.

pisek/env/task_config.py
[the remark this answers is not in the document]
SvizelPritula: Ok, that makes sense.

pisek/visualize.py
SvizelPritula: Whoops, I didn't notice that.

pisek/visualize.py
SvizelPritula: An item can appear in a nested list multiple times.
Since this is a dictionary that uses numbers from `0` to `n-1` as keys, it could be a `list[list[LoggedResult]]` instead.
It is true that this matches how subtasks are stored inside of `TaskConfig`, so it might be best to keep it as is.

5K1PY: Ok, changed. I don't think it matters that much and we can change it later.

pisek/jobs/parts/verdicts_eval.py
```diff
@@ -0,0 +1,30 @@
...
+        definitive &= (
+            (quant == any and ok)
+            or (quant == all and not ok)
+            or (SUBTASK_SPEC[expected][i] == verdict_always)
+        )
+        if quant == all and ok == False:
+            breaker = oks.index(False)
+            break
+
+    return result, definitive, breaker
```
SvizelPritula, line 30: > Given how different `any` and `all` behave, wouldn't it be easier to just add an `if` here?
> You could also move `ok = quant(oks)` into the cases and avoid passing functions around entirely.

> I don't see how it simplifies things.

It could be written like this:

```suggestion
def evaluate_verdicts(
    config: TaskConfig, verdicts: list[Verdict], expected: str
) -> tuple[bool, bool, Optional[int]]:
    result = True
    definitive = True
    breaker = None

    modes = [FailMode.all, config.fail_mode]

    for i, mode in enumerate(modes):
        oks = list(map(SUBTASK_SPEC[expected][i], verdicts))

        if mode == FailMode.all:
            ok = all(oks)

            result &= ok
            definitive &= not ok or SUBTASK_SPEC[expected][i] == verdict_always

            if not ok:
                breaker = oks.index(False)
                break
        elif mode == FailMode.any:
            ok = any(oks)

            result &= ok
            definitive &= ok

    return result, definitive, breaker
```

This would make it clearer how each mode is checked. One downside is that `result &= ok` is duplicated.

SvizelPritula (approved):
LGTM

PR #153, opened 2024-03-10
Config errors

### LINKED ISSUE #146: Better config errors
Currently validation errors of config have to do some automagic to guess where the variable came from. Let's do something more systematic.

5K1PY:
Fixes #146


pisek/env/config_hierarchy.py
```diff
@@ -16,27 +16,47 @@
...
+@dataclass
+class ConfigValue:
+    value: Any
+    config: str
+    section: str
+    key: Optional[str]
+    internal: bool = False
+
+    def location(self) -> str:
+        text = "Internal pisek value" if self.internal else f"Config {self.config}"
```
SvizelPritula, line 42: I find it a bit confusing how the message just starts with "Config config".
Could we maybe force it to say "Config ./config"?

pisek/env/task_config.py
```diff
@@ -612,12 +626,23 @@ def _format_message(err: ErrorDetails) -> str:
...
+        if not isinstance(value, ConfigValue):
+            location = (
+                value["_section"].location()
+                if "_section" in value
+                else "Global config error:"
+            )
+        else:
+            location = value.location()
+
+        error_msgs.append(f"{location}\n" + tab(_format_message(error)))
```
SvizelPritula, line 645: Wouldn't it be clearer if the message looked like this?
```suggestion
        error_msgs.append(f"In {location}:\n" + tab(_format_message(error)))
```

With `location()` changed to output "config..." instead of "Config...".

SvizelPritula (changes requested):
There are a couple of issues.

pisek/env/task_config.py
5K1PY: Changed [here](https://github.com/kasiopea-org/pisek/pull/153/files#diff-32f098dcd2b4f7a414a0bba8124f892c4eb4d19a175f845fac8965035032f9b8R42).

SvizelPritula (changes requested):

pisek/env/config_hierarchy.py
```diff
@@ -16,27 +16,49 @@
...
+    value: Any
+    config: str
+    section: str
+    key: Optional[str]
+    internal: bool = False
+
+    def location(self) -> str:
+        text = (
+            "Internal pisek value" if self.internal else f"In section [{self.section}]"
+        )
```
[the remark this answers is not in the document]
5K1PY: oops thanks for caching that

SvizelPritula (approved):
LGTM

PR #156, opened 2024-03-14
Verbose mode

### LINKED ISSUE #155: Verbose mode
Add verbose mode that will make `SolutionManager`'s output more granular:
- When `verbosity=0` keep as is
- When `verbosity=1` print details per subtasks
- When `verbosity=2` print details per input

Also we could add to output slowest time per solution / subtask / in test.

5K1PY:
Fixes #155

### DIFF
--- a/pisek/__main__.py
+++ b/pisek/__main__.py
@@ -78,9 +78,9 @@ def main(argv):
         parser.add_argument(
             "--verbosity",
             "-v",
-            type=int,
+            action="count",
             default=0,
-            help="Be more verbose.",  # TODO
+            help="Be more verbose. (Enter multiple times for even more verbosity.)",
         )
 
     def add_argument_timeout(parser):
--- a/pisek/__main__.py
+++ b/pisek/__main__.py
@@ -74,6 +74,15 @@ def main(argv):
         )
     )
 
+    def add_argument_verbosity(parser):
+        parser.add_argument(
+            "--verbosity",
+            "-v",
+            type=int,
+            default=0,
+            help="Be more verbose.",  # TODO
+        )
+
     def add_argument_timeout(parser):
         parser.add_argument(
             "--timeout",
@@ -223,6 +232,7 @@ def main(argv):
             help="Use the dataset with the description DESCRIPTION.",
         )
 
+    add_argument_verbosity(parser)
     add_argument_timeout(parser)
     add_argument_full(parser)
     add_argument_strict(parser)
--- a/pisek/jobs/parts/solution.py
+++ b/pisek/jobs/parts/solution.py
@@ -139,21 +139,23 @@ class SolutionManager(TaskJobManager):
 
         max_time = max((s.slowest_time for s in self.subtasks), default=0)
 
-        verbosity = self._env.verbosity if self.state.finished() else 0
-        subtasks_res = [sub.status(self.subtasks, verbosity) for sub in self.subtasks]
-
-        if verbosity <= 0:
+        if not self.state.finished() or self._env.verbosity == 0:
             points = pad_left(points, points_places)
             header = f"{pad(msg, MSG_LEN-1)} {points}   "
-            subtasks_text = "|".join(subtasks_res)
+            subtasks_text = "|".join(sub.status_verbosity0() for sub in self.subtasks)
         else:
             header = (
                 right_aligned_text(f"{msg}: {points}", f"slowest {max_time:.2f}s")
                 + "\n"
             )
             header = colored_env(header, "cyan", self._env)
-            subtasks_text = tab("\n".join(subtasks_res))
-            if verbosity == 1:
+            subtasks_text = tab(
+                "\n".join(
+                    sub.status(self.subtasks, self._env.verbosity)
+                    for sub in self.subtasks
+                )
+            )
+            if self._env.verbosity == 1:
                 subtasks_text += "\n"
 
         return header + subtasks_text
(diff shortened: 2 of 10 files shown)


pisek/jobs/parts/solution.py
```diff
@@ -229,31 +256,99 @@ def _jobs_points(self) -> list[float]:
...
+
+            text += right_aligned_text(
+                f"{subtask_name} {subtask_points}p  "
+                f"{predecessor_summary}{verdict_marks(self.new_jobs)} ",
+                f"slowest {self.slowest_time:.2f}s",
+                offset=-2,
+            )
+
+        elif verbosity >= 2:
+            longest_input_name = max(map(lambda j: len(j.input.name), self.new_jobs))
```
SvizelPritula, line 306: Nit: Why does this use `map` when `max_sub_name_len` and `max_sub_points_len` use list comprehension? Why is the naming scheme completely different?

pisek/jobs/parts/solution.py
```diff
@@ -229,31 +256,99 @@ def _jobs_points(self) -> list[float]:
...
+            if len(self.subtask.predecessors) and any(
+                map(lambda j: j.result, self.new_jobs)
+            ):
+                text += "\n"
+
+            for job in self.new_jobs:
+                if job.result is not None:
+                    input_verdict = tab(
+                        f"{pad(job.input.name + ':', longest_input_name+1)} "
+                        f"{pad(job.verdict_text(), Verdict.pad_length())} "
```
SvizelPritula, line 342: Is there a reason this needs to be padded? `right_aligned_text` will add spaces anyways.

pisek/jobs/parts/solution.py
```diff
@@ -229,31 +256,99 @@ def _jobs_points(self) -> list[float]:
...
+                    )
+                    text += (
+                        right_aligned_text(
+                            input_verdict, f"{job.result.time:.2f}s", offset=-2
+                        )
+                        + "\n"
+                    )
```
SvizelPritula, line 349: Nit: I think this would be cleaner:
```suggestion
                    text += right_aligned_text(
                        input_verdict, f"{job.result.time:.2f}s", offset=-2
                    )
                    text += "\n"
```

pisek/utils/terminal.py
```diff
@@ -25,6 +25,12 @@
 except OSError:
     terminal_width, terminal_height = 100, 24
 
+TARGET_LINE_WIDTH = min(terminal_width, 100)
+
+
+def right_aligned_text(left: str, right: str, offset: int = 0):
```
SvizelPritula, line 31: Nit: Is this a good name? It doesn't *just* right align text, it also left aligns some.

pisek/jobs/parts/solution.py
```diff
@@ -229,31 +256,99 @@ def _jobs_points(self) -> list[float]:
...
+            return "".join(job.verdict_mark() for job in jobs)
+
+        verbosity = self._env.verbosity if verbosity is None else verbosity
+        text = ""
+
+        predecessor_summary = verdict_summary(self.previous_jobs)
+        if predecessor_summary:
+            predecessor_summary = f"({predecessor_summary}) "
+
+        if verbosity == 0:
```
SvizelPritula, line 284: This doesn't handle negative verbosities.

pisek/jobs/parts/solution.py
```diff
@@ -229,31 +256,99 @@ def _jobs_points(self) -> list[float]:
             )
         )
 
-    def __str__(self) -> str:
-        s = "("
-        previous = list(
-            map(
-                lambda x: x.verdict if x else None,
-                self._job_results(self.previous_jobs),
+    def status(
```
SvizelPritula, line 259: Woudn't it be more readable if this were multiple methods split by verbosity? All the different variants share is code to wrap the predecessor summary in parentheses.
This would also avoid the awkwardness needed when calling this method to determine whether the resulting text should be displayed inline or on multiple lines.

pisek/jobs/parts/solution.py
```diff
@@ -130,31 +130,41 @@ def _update(self):
...
+        points_places = len(str(self._env.config.total_points)) + POINTS_DEC_PLACES + 2
+        points = f"{format_points(self.solution_points)}p"
+
+        max_time = max((s.slowest_time for s in self.subtasks), default=0)
+
+        verbosity = self._env.verbosity if self.state.finished() else 0
+        subtasks_res = [sub.status(self.subtasks, verbosity) for sub in self.subtasks]
+
+        if verbosity <= 0:
+            points = pad_left(points, points_places + 1)
```
SvizelPritula, line 146: Why add 1 here? Wouldn't it be clearer if the space was added in the format string a line below?

pisek/jobs/parts/solution.py
```diff
@@ -130,31 +130,41 @@ def _update(self):
...
+            header = f"{pad(msg, MSG_LEN)}{points} "
+            subtasks_text = tab("|".join(subtasks_res))
```
SvizelPritula, line 148: This adds three spaces beween the header and subtask text in total: One from the format string, and two from `tab`. This is also a really awkward use of tab in the first place.

pisek/utils/terminal.py
```diff
@@ -25,6 +25,12 @@
...
+    return pad(left, TARGET_LINE_WIDTH - len(right) + offset) + right
```
SvizelPritula, line 32: Woudn't it better if this function ensured there was at least one space between `left` and `right`?
As far as I can tell, this needs to be done manually at all call sites.

pisek/__main__.py
```diff
@@ -74,6 +74,15 @@ def main(argv):
...
     )
 
+    def add_argument_verbosity(parser):
+        parser.add_argument(
+            "--verbosity",
+            "-v",
+            type=int,
+            default=0,
+            help="Be more verbose.",  # TODO
+        )
```
SvizelPritula, line 84: Idea: In most commands, increasing verbosity further is done by using `-v` multiple times. Do we want that? `argparse` supports it with `action='count'`.

SvizelPritula (changes requested):
There are a couple of issues.

pisek/jobs/status.py
```diff
@@ -29,7 +28,7 @@ def _bar(self, msg: str, part: int, full: int, color: str = "cyan") -> str:
         msg = pad(msg, MSG_LEN)
         progress_msg = f"  ({part}/{full})"
 
-        bar_len = min(terminal_width - len(msg) - len(progress_msg), MAX_BAR_LEN)
+        bar_len = TARGET_LINE_WIDTH - MSG_LEN - len(progress_msg)
```
SvizelPritula, line 31: This has the disadvantage of having the bar length change as `part` increases, as well as nearly always having different bars have different lengths. The original logic wasn't *bad*, it was just confusingly written.
How about something like this?

```
TARGET_BAR_WIDTH = TARGET_LINE_WIDTH - MSG_LEN - 11
# ...
bar_len = min(TARGET_BAR_WIDTH, TARGET_LINE_WIDTH - len(msg) - len(progress_msg))
```

Maybe it would be a good idea to add a minimum length as well, say `10` characters.

pisek/utils/terminal.py
SvizelPritula: I'm not sure `split_aligned_text` is that great, either, as it (at least to me) sounds like it splits something.
I can't really think of a good name for the function. Maybe `right_aligned_text` is the best option after all?

SvizelPritula (changes requested):

pisek/utils/terminal.py
SvizelPritula: Great!
However, a space is still added manually at most callsites, which should be removed.

pisek/jobs/status.py
```diff
@@ -29,7 +28,10 @@ def _bar(self, msg: str, part: int, full: int, color: str = "cyan") -> str:
         msg = pad(msg, MSG_LEN)
         progress_msg = f"  ({part}/{full})"
 
-        bar_len = min(terminal_width - len(msg) - len(progress_msg), MAX_BAR_LEN)
+        TARGET_BAR_WIDTH = TARGET_LINE_WIDTH - MSG_LEN - 11
```
SvizelPritula, line 31: It's a bit weird to declare a constant inside of a function. (See also [PEP 8](https://peps.python.org/pep-0008/#constants).)

SvizelPritula (changes requested):
It looks good, but I maintain that the `SubtaskJobGroup.status` method is too long and should be split up by verbosity.
This would also make it easier to override the verbosity when the job hasn't finished yet.

SvizelPritula (approved):
LGTM

PR #158, opened 2024-04-05
Improve message on solution resulting unexpectedly

### LINKED ISSUE #148: Neukazování diffů, když je použitý judge
Ukazování diffu u úloh, kde je použitý judge je dost matoucí.

image

Vypisovaný diff nemusí být vůbec zajímavý a spíše zmate člověka a myslí si, že úloha je kontrolována pomocí diffu.
Nejvíce matoucí je to v době, když se úloha během vývoje mění z diffu na judge, tak z této hlášky to není vůbec poznat.

5K1PY:
Makes reports more reasonable.

Fixes #148, #149

### DIFF
--- a/pisek/jobs/parts/judge.py
+++ b/pisek/jobs/parts/judge.py
@@ -250,12 +250,18 @@ class RunJudge(ProgramsJob):
         elif self.result.verdict == Verdict.timeout:
             head = f"Solution did timeout on input {self.input}"
 
-        # text = f"{head}:\n{tab(self.result.output)}"
+        text = f"{head}:\n"
+        for prog in ("solution", "judge"):
+            rr = getattr(self.result, f"{prog}_rr")
+            if rr is not None:
+                text += (
+                    tab(f"{prog.capitalize()}:\n" + tab(self._format_run_result(rr))) + "\n"
+                )
         # TODO: Diff
         # if self.result.diff != "":
         #     text += "\n" + tab(f"diff:\n{tab(self.result.diff)}")
 
-        return head
+        return text
 
     def verdict_text(self) -> str:
         if self.result is not None:
--- a/pisek/jobs/parts/judge.py
+++ b/pisek/jobs/parts/judge.py
@@ -246,20 +246,27 @@ class RunJudge(ProgramsJob):
             text += f"correct output: {self._named_file(self.correct_output)}"
         text += f"result: {self.result.verdict.name}\n"
 
-        text += f"solution:\n"
-        text += tab(
-            self._format_run_result(
-                sol_rr,
-                status=sol_rr.kind != RunResultKind.OK,
-                stderr_force=sol_rr.kind == RunResultKind.RUNTIME_ERROR,
-                time=True
+        if isinstance(self, RunBatchJudge):
+            text += f"solution:\n"
+            text += tab(
+                self._format_run_result(
+                    sol_rr,
+                    status=sol_rr.kind != RunResultKind.OK,
+                    stderr_force=sol_rr.kind == RunResultKind.RUNTIME_ERROR,
+                    time=True,
+                )
             )
-        )
-        text += "\n"
+            text += "\n"
         if judge_rr is not None:
             text += (
                 f"{self.judge_name}:\n"
-                + tab(self._format_run_result(judge_rr, status=False, stderr_force=True))
+                + tab(
+                    self._format_run_result(
+                        judge_rr,
+                        status=isinstance(self, RunDiffJudge),
+                        stderr_force=True,
+                    )
+                )
                 + "\n"
             )
 
--- a/pisek/jobs/parts/judge.py
+++ b/pisek/jobs/parts/judge.py
@@ -218,19 +218,9 @@ class RunJudge(ProgramsJob):
         if self._solution_run_res.kind == RunResultKind.OK:
             result = self._judge()
         elif self._solution_run_res.kind == RunResultKind.RUNTIME_ERROR:
-            result = SolutionResult(
-                Verdict.error,
-                0.0,
-                self._solution_run_res,
-                None
-            )
+            result = SolutionResult(Verdict.error, 0.0, self._solution_run_res, None)
         elif self._solution_run_res.kind == RunResultKind.TIMEOUT:
-            result = SolutionResult(
-                Verdict.timeout,
-                0.0,
-                self._solution_run_res,
-                None
-            )
+            result = SolutionResult(Verdict.timeout, 0.0, self._solution_run_res, None)
 
         if (
             self.expected_points is not None
@@ -327,10 +317,7 @@ class RunCMSJudge(RunJudge):
                 verdict = Verdict.partial
 
             return SolutionResult(
-                verdict,
-                points,
-                self._solution_run_res,
-                judge_run_result
+                verdict, points, self._solution_run_res, judge_run_result
             )
         else:
             raise self._create_program_failure(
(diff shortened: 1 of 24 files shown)


pisek/jobs/parts/task_job.py
```diff
@@ -183,3 +190,25 @@ def _diff_files(self, file_a: TaskPath, file_b: TaskPath) -> str:
             stdout=subprocess.PIPE,
         )
         return diff.stdout.decode("utf-8")
+
+    def _file_content(
```
SvizelPritula, line 194: This method name sounds like it returns the entire file. Wouldn't something like `_quote_file` be better?

pisek/jobs/parts/task_job.py
```diff
@@ -183,3 +190,25 @@ def _diff_files(self, file_a: TaskPath, file_b: TaskPath) -> str:
...
+                text = colored_env(text, color, self._env)
+                for _ in range(tab_times):
+                    text = tab(text)
+            return text + "\n"
+        else:
+            self._access_file(file)  # Caching improvements
+            return ""
+
+    def _named_file(self, file: TaskPath, **kwargs) -> str:
+        return f"{file.col(self._env)}\n{self._file_content(file, **kwargs)}"
```
SvizelPritula, line 214: This function name seems really unclear, since every file is named. How about `_quote_file_with_name` or someting similar?

pisek/jobs/parts/judge.py
```diff
@@ -417,28 +417,13 @@ def _judge(self) -> SolutionResult:
...
+            return SolutionResult(Verdict.ok, 1.0, self._solution_run_res, rr)
         elif diff.returncode == 1:
-            return SolutionResult(
-                Verdict.wrong_answer,
-                0.0,
-                self._solution_run_res.time,
-                self._solution_run_res.wall_time,
-                "",
-                self._quote_program(rr),
-                self._nice_diff(),
```
SvizelPritula, line 440: `_nice_diff` isn't used anywhere anymore. Should we remove it?

pisek/jobs/parts/task_job.py
```diff
@@ -183,3 +190,25 @@ def _diff_files(self, file_a: TaskPath, file_b: TaskPath) -> str:
             stdout=subprocess.PIPE,
         )
         return diff.stdout.decode("utf-8")
+
+    def _file_content(
+        self,
+        file: TaskPath,
+        force: bool = False,
+        tab_times: int = 1,
```
SvizelPritula, line 198: What is the intended usecase for this? The convention seems to be that indenting text is the responsibility of whoever combines it with other text. I don't think this function should indent the result at all, that should be left to `_named_file`.

pisek/jobs/parts/task_job.py
```diff
@@ -183,3 +190,25 @@ def _diff_files(self, file_a: TaskPath, file_b: TaskPath) -> str:
             stdout=subprocess.PIPE,
         )
         return diff.stdout.decode("utf-8")
+
+    def _file_content(
+        self,
+        file: TaskPath,
+        force: bool = False,
```
SvizelPritula, line 197: Here it's clear what `force` means, but not when it's forwarded from another function.

SvizelPritula (changes requested):
Looks great! There are a few issues with the implementation.

pisek/jobs/parts/judge.py
```diff
@@ -250,27 +238,43 @@ def message(self) -> str:
...
-            text += "\n" + tab(f"diff:\n{tab(self.result.diff)}")
+        sol_rr = self.result.solution_rr
+        judge_rr = self.result.judge_rr
+
+        text = f"input: {self._named_file(self.input)}"
+        if isinstance(self, RunBatchJudge):
+            text += f"correct output: {self._named_file(self.correct_output)}"
+        text += f"result: {self.result.verdict.name}\n"
+
+        if isinstance(self, RunBatchJudge):
```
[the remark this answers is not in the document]
5K1PY: Well, only batch judges have correct output. There is no such thing for communication.

pisek/jobs/parts/run_result.py
```diff
@@ -0,0 +1,80 @@
...
+    that finished successfully, but got Wrong Answer, still gets the OK
+    RunResult."""
+
+    def __init__(
+        self,
+        kind: RunResultKind,
+        returncode: int,
+        time: float,
+        wall_time: float,
+        stdout_file: Optional[Union[TaskPath, int]] = None,
```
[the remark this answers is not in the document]
5K1PY: When stdout is pipe. Why should ``_named_file`` handle that?

pisek/jobs/parts/judge.py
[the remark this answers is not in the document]
SvizelPritula: That's true, but they do have a solution run result. It's impossible to print their stdout, but they still have an interesting stderr and time.

pisek/jobs/parts/run_result.py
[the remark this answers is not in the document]
SvizelPritula: In that case, we should make sure we never try to print the contents of a pipe. Currently that works becouse nothing gets printed about communication solutions, but I would consider that a mistake.

pisek/jobs/parts/task_job.py
5K1PY: This gets very annoying if we are actualy not printing anything when ``env.file_contents == False``.

SvizelPritula: What do you mean? It's true that you would have to add an `if` to `_named_file`, but I don't see how that's more annoying than adding a for loop here. I'm also unsure if this function shold be responsible for checking the `env`. Shouldn't that also be done in `_named_file`?

pisek/jobs/parts/run_result.py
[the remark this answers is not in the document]
5K1PY: > In that case, we should make sure we never try to print the contents of a pipe. Currently that works becouse nothing gets printed about communication solutions, but I would consider that a mistake.

Yep, we are doing that in `_format_run_result`.

pisek/jobs/parts/task_job.py
```diff
@@ -77,18 +78,24 @@ def globs_to_files(
...
+            line = line.strip()
+            if len(line) > max_chars:
+                line = line[: max_chars - 1] + "…"
+            s_text.append(line)
+        if len(s_text) < max_lines:
+            return text
+
+        tail = max(ceil(("t" in style) * max_lines / len(style)) - 1, 0)
+        head = max_lines - tail - 1
+        return "\n".join(s_text[:head] + ["[...]"] + (s_text[-tail:] if tail else []))
```
SvizelPritula, line 98: We should probably also use the unicode ellipsis character here?

```suggestion
        return "\n".join(s_text[:head] + ["[…]"] + (s_text[-tail:] if tail else []))
```

pisek/jobs/parts/task_job.py
```diff
@@ -77,18 +78,24 @@ def globs_to_files(
...
+        max_chars: int = 100,
+    ) -> str:
+        s_text = []
+        for line in text.split("\n"):
+            line = line.strip()
+            if len(line) > max_chars:
+                line = line[: max_chars - 1] + "…"
+            s_text.append(line)
+        if len(s_text) < max_lines:
+            return text
```
SvizelPritula, line 94: If the file has less than `max_lines` lines, lines won't be truncated.

```suggestion
            return "\n".join(s_text)
```

SvizelPritula (changes requested):

SvizelPritula (approved):
LGTM

PR #161, opened 2024-04-09
Generate testing log even if solution fails

### LINKED ISSUE #152: Testing log isn't generated in case of failing solution
Any error during testing will prevent a testing log from being generated.
This effectively means `visualize` cannot be used unless you pick a valid time limit beforehand, which makes it effectively useless.

5K1PY:
Generalized job core for `Job`s running always. `JobManger`s have to generate result even if `JobManager._evaluate` fails.

Fixes #152.

### DIFF
--- a/pisek/jobs/jobs.py
+++ b/pisek/jobs/jobs.py
@@ -72,6 +72,7 @@ class CaptureInitParams:
 class PipelineItem(ABC):
     """Generic PipelineItem with state and dependencies."""
 
+    _env: Env
     run_always: bool = False  # Runs even if prerequisites failed
 
     def __init__(self, name: str) -> None:
@@ -362,7 +363,7 @@ class JobManager(PipelineItem):
             else:
                 self.state = State.succeeded
 
-            self.result = self._compute_result()
+        self.result = self._compute_result()
 
         super().finish()
         return self._get_status()
--- a/pisek/jobs/jobs.py
+++ b/pisek/jobs/jobs.py
@@ -27,6 +27,7 @@ import os.path
 from pisek.jobs.cache import Cache, CacheEntry
 from pisek.env.env import Env
 from pisek.utils.paths import TaskPath
+from pisek.utils.terminal import colored_env
 
 
 class State(Enum):
@@ -88,6 +89,12 @@ class PipelineItem(ABC):
         """Prints text to stdout/stderr."""
         self.dirty = True
         (sys.stderr if stderr else sys.stdout).write(msg + end)
+
+    def _warn(self, msg: str) -> None:
+        if self._env.strict:
+            raise PipelineItemFailure(msg)
+        else:
+            self._print(colored_env(msg, "yellow", self._env))
 
     def _fail(self, failure: PipelineItemFailure) -> None:
         """End this job in failure."""
--- a/pisek/jobs/jobs.py
+++ b/pisek/jobs/jobs.py
@@ -71,6 +71,8 @@ class CaptureInitParams:
 class PipelineItem(ABC):
     """Generic PipelineItem with state and dependencies."""
 
+    run_always: bool = False  # Runs even if prerequisites failed
+
     def __init__(self, name: str) -> None:
         self.name = name
         self.state = State.in_queue
@@ -100,7 +102,8 @@ class PipelineItem(ABC):
             return  # No need to cancel
         self.state = State.canceled
         for item, _ in self.required_by:
-            item.cancel()
+            if not item.run_always:
+                item.cancel()
 
     def _check_prerequisites(self) -> None:
         """Checks if all prerequisites are finished raises error otherwise."""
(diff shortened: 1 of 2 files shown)


pisek/jobs/jobs.py
```diff
@@ -100,11 +102,12 @@ def cancel(self) -> None:
...
         self.state = State.canceled
         for item, _ in self.required_by:
-            item.cancel()
+            if not item.run_always:
+                item.cancel()
 
     def _check_prerequisites(self) -> None:
         """Checks if all prerequisites are finished raises error otherwise."""
-        if self.prerequisites > 0:
+        if not self.run_always and self.prerequisites > 0:
```
SvizelPritula, line 110: I'll admit I don't know a lot about the inner workings of the pipeline, but isn't this redundant with the changes to `finish`?
`finish` has been changed to decrement `prerequisites` for `run_always` jobs, so I'd assume bypassing the `prerequisites > 0` check is unnecessary?

pisek/jobs/parts/testing_log.py
```diff
@@ -61,12 +65,17 @@ def _evaluate(self) -> None:
                     }
                 )
 
-        if warn_skipped:
-            MSG = "Not all inputs were tested. For testing them use --all-inputs."
+        warn_msg = None
+        if len(solutions) < len(self._env.solutions):
+            warn_msg = "Not all solutions were tested."
+        elif warn_skipped:
+            warn_msg = "Not all inputs were tested. For testing them use --all-inputs."
```
SvizelPritula, line 72: It would be nice to print both warnings, at least if `strict` is not set.

SvizelPritula:
A testing log also doesn't get generated if a solution fails unless `--full` is passed. I haven't managed to get the "Not all solutions were tested." warning to appear.

SvizelPritula:
If you request a testing log when a solution doesn't compile, you gen an exception:

```
Traceback (most recent call last):
  File "/home/dev/.local/bin/pisek", line 8, in <module>
    sys.exit(main_wrapped())
             ^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/__main__.py", line 398, in main_wrapped
    result = main(sys.argv[1:])
             ^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/__main__.py", line 359, in main
    result = test_task(args, solutions=None, target="all")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/utils/pipeline_tools.py", line 77, in g
    res = f(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/__main__.py", line 41, in test_task
    return test_task_path(PATH, **vars(args), **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/__main__.py", line 45, in test_task_path
    return run_pipeline(path, TaskPipeline, solutions=solutions, **env_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/utils/pipeline_tools.py", line 40, in run_pipeline
    return pipeline.run_jobs(Cache(env), env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/jobs/job_pipeline.py", line 60, in run_jobs
    self.failed |= not self._status_update(env.fork())
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/jobs/job_pipeline.py", line 83, in _status_update
    msg = job_man.finalize()
          ^^^^^^^^^^^^^^^^^^
  File "/home/dev/Code/pisek/pisek/jobs/jobs.py", line 352, in finalize
    self._evaluate()
  File "/home/dev/Code/pisek/pisek/jobs/parts/testing_log.py", line 54, in _evaluate
    for inp, sol_res in data["results"].items():
                        ~~~~^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

pisek/jobs/jobs.py
5K1PY: For example when generator fails, testing solutions is canceled and therefore ``finish`` is not called on them. Therefore they do not substract anything from ``self.prerequisites``.

5K1PY:
> A testing log also doesn't get generated if a solution fails unless `--full` is passed.

That's intentional. I think we should stick to failing fast in all cases unless `--full` is provided.

> I haven't managed to get the "Not all solutions were tested." warning to appear.

One case is if generator fails. I have added also when testing single solution.

pisek/jobs/jobs.py
SvizelPritula: In that case, why are the changes in `finish` needed?

SvizelPritula (approved):
LGTM

PR #164, opened 2024-04-19
Config v3

### LINKED ISSUE #133: Replace clock_limit with automatic setting
``clock_limit`` can be confusing when setting ``time_limit`` higher.

Replacing it with ``clock_mul`` a ``clock_min`` and setting clock limit to:
```
max(clock_mul * time_limit, clock_min)
```
would work

### LINKED ISSUE #135: Hierarchical versions of config

5K1PY:
Adds expermental config version 3.

* Supports hierarchical configs for defaults (Fixes #135)
* Replaces clock_limit with automatic setting (Fixes #133)
* Allows data size checking for CMS
* Removes ignored keys


pisek/config/config_hierarchy.py
```diff
@@ -128,27 +185,16 @@ def check_unused_keys(self) -> None:
...
                 if key in IGNORED_KEYS[section]:
                     continue
-                if (
-                    section == "all_tests" or re.match(r"test\d{2}", section)
-                ) and key in IGNORED_TEST_KEYS:
-                    continue
                 if key not in self._used_keys[section]:
                     raise TaskConfigError(
-                        f"Unexpected key '{key}' in section [{section}] of config."
+                        f"Unexpected key '{key}' in section [{section}] of config. {self._help_invalid_key(section, key)}"
```
SvizelPritula, line 197: Nit: If there is no help, there will be a trailing space.

pisek/config/config_hierarchy.py
```diff
@@ -159,3 +205,30 @@ def check_todos(self) -> bool:
...
+            f"(Did you mean '{guess[1]}'"
+            + (f" in section [{guess[0]}]" if section != guess[0] else "")
+            + "?)"
+        )
+
+    def _guess_key(self, section: str, key: str) -> Optional[tuple[str, str]]:
+        v3_docs = ConfigParser()
+        v3_docs.read(V3_DOCUMENTATION)
+
+        best_guess = ("", "")
```
SvizelPritula, line 224: Wouldn't it be much cleaner to use `None` to mark no match?

pisek/config/config_types.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class JudgeType(StrEnum):
+    diff = auto()
+    judge = auto()
+
+
+class Scoring(StrEnum):
+    equal = auto()
+    min = auto()
```
SvizelPritula, line 28: These names aren't very intuitive.

pisek/config/global-defaults
```diff
@@ -34,32 +33,38 @@ subtasks=@auto
 
 [limits]
 tool_time_limit=0
-tool_clock_limit=0
+tool_clock_mul=2
+tool_clock_min=1
```
SvizelPritula, line 37: This also isn't mentioned in the example config.

pisek/config/task_config.py
```diff
@@ -507,11 +472,18 @@ class ProgramLimits(BaseEnv):
     _section: str = "limits"
 
     time_limit: float = Field(ge=0)  # [seconds]
-    clock_limit: float = Field(ge=0)  # [seconds]
+    clock_mul: float = Field(ge=0)  # [1]
+    clock_min: float = Field(ge=0)  # [seconds]
```
SvizelPritula, line 476: Nit: Note that this still doesn't allow us to match CMS's behaviour exactly, as it computes the clock limit as `time_limit * 2 + 1`

pisek/config/update_config.py
```diff
@@ -162,24 +168,66 @@ def get_subtask_mask(points, subtasks):
...
+            "limits", f"{program_type}_clock_limit", fallback="360"
+        )
+        maybe_delete_key(config, "limits", f"{program_type}_clock_limit")
+
+    IGNORED_KEYS = [
+        ("task", "tests"),
+        ("tests", "in_mode"),
+        ("tests", "out_mode"),
+        ("tests", "out_format"),
+        ("tests", "online_validity"),
```
SvizelPritula, line 190: We'll probably want to support these two keys at some point.

pisek/config/update_config.py
```diff
@@ -162,24 +168,66 @@ def get_subtask_mask(points, subtasks):
...
+    for section in config.sections():
+        if re.match(r"test\d{2}", section):
+            maybe_delete_key(config, section, "file_name")
+
+
+OUTDATED_VERSIONS = {"v1": ("v2", update_to_v2)}
+CURRENT_VERSIONS = {"v2": ("v3", update_to_v3)}
+
+NEWEST_VERSION = "v3"
+IS_EXPERIMENTAL = True
```
SvizelPritula, line 204: Nit: It's not completely clear what's experimental.

tests/test_kasiopea.py
```diff
@@ -217,17 +217,6 @@ def modify_task(self):
...
-    def expecting_success(self):
-        return False
-
-    def modify_task(self):
-        def modification_fn(raw_config):
-            raw_config["limits"]["input_max_size"] = "0"
-
-        modify_config(self.task_dir, modification_fn)
-
-
```
SvizelPritula, line 230: Is there a reason we can't keep these configs?

SvizelPritula (changes requested):
There are a few oddities.

pisek/config/config_types.py
5K1PY: What do you suggest?

SvizelPritula: I don't know. I feel like `all` was a bit more intuitive, but I'm not sure it that's that great either.

pisek/config/task_config.py
5K1PY: I know. But:
1) Your solutions shouldn't timeout on wall clock anyway.
2) I kind of find this solution more intuitive
3) Most of the times you should retest it on CMS anyway.

pisek/config/update_config.py
5K1PY: Lucky the v3 is experimental and we can change this code :innocent:

tests/test_kasiopea.py
5K1PY: I don't know about these. Cause now zero means unlimited testing this would probably mean stuffing the inputs with whitespace, which I don't particularly like. I am thinking more about adding new fixtures and keeping these for testing v2->v3 conversion. What do you think?

SvizelPritula: That's fair, all of our fixtures have inputs and outputs less than a megabyte in size.
I guess we could modify `odd_stub` to generate bigger inputs pretty easily, through.

pisek/config/update_config.py
SvizelPritula: Given that updating is currently in memory only, I guess there also is no possibility of anyone accidentally deleting data from their config file, so I guess it's fine.

pisek/config/task_config.py
SvizelPritula: Fair.

config-v3-documentation
```diff
@@ -158,14 +170,16 @@ subtasks=@auto
...
 #   in_gen      Input generator
 #   checker     Checker
 #   judge       Judge
 #   solve       Primary solution
 #   sec_solve   Secondary solution usually subject to much less strict limits.
 # and LIMIT is one of:
 #   time_limit  Execution time limit [seconds]
-#   clock_limit Wall clock time limit [seconds]
+#   clock_mul   Wall clock multiplier [1]
+#   clock_min   Wall clock minimum [second]
```
SvizelPritula, line 182: It isn't clear what the multiplier multiplies.

```suggestion
#   clock_mul   Wall clock time limit, as a multiple of the time limit
#   clock_min   Wall clock time limit minimum, regardless of the time limit [second]   
```

SvizelPritula (changes requested):
There are still a couple of tiny issues.

config-v3-documentation
5K1PY: Is not the description on lines 191-192 not enogh?

Also it's not regardles because zero behaves wierd.

SvizelPritula: Sorry, I missed that.

tests/test_kasiopea.py
5K1PY: Done

PR #165, opened 2024-05-02
Judge type

5K1PY:
Separated judge calling and judging convention from ``contest_type``.

### DIFF
--- a/config-v3-documentation
+++ b/config-v3-documentation
@@ -69,6 +69,16 @@ out_judge=judge
 # Only for out_check=judge
 # Source code of judge (without suffix)
 
+judge_type=cms
+# Specifies how to call judge and how judge reports result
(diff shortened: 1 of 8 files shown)


SvizelPritula (commented):
LGTM

We'll probably want to implement support for the judge format used by KSP, which is identical to the Kasiopea format, but uses exit codes 42 and 43 instead of 0 and 1. It might be worth considering having both use the same `judge_type`, but add keys that control the meaning of exit codes.

5K1PY:
> We'll probably want to implement support for the judge format used by KSP, which is identical to the Kasiopea format, but uses exit codes 42 and 43 instead of 0 and 1. It might be worth considering having both use the same judge_type, but add keys that control the meaning of exit codes.

But we don't want to make 65280 standards... I will implement 42/43 later

PR #174, opened 2024-05-13
DataChecking: Support in_format and out_format keys

### LINKED ISSUE #169: Binary input
Add following config options 
```
in_format=text
out_format=binary
```
Where binary input / output is not checked for cleanliness.

SvizelPritula:
Adds support for the `in_format` and `out_format` keys, which can be set to `text` or `binary`. Setting the correct key to `binary` will stop pisek from checking if the relevant files are UNIX text files and contain valid ASCII chars.

Resolves #169.

### DIFF
--- a/pisek/config/config_types.py
+++ b/pisek/config/config_types.py
@@ -28,6 +28,11 @@ class JudgeType(StrEnum):
     opendata_v1 = "opendata-v1"
 
 
+class FormatType(StrEnum):
+    text = auto()
+    binary = auto()
+
+
 class Scoring(StrEnum):
     equal = auto()
     min = auto()
--- a/pisek/config/config_types.py
+++ b/pisek/config/config_types.py
@@ -28,7 +28,7 @@ class JudgeType(StrEnum):
     opendata_v1 = "opendata-v1"
 
 
-class FormatType(StrEnum):
+class DataFormat(StrEnum):
     text = auto()
     binary = auto()
 
--- a/tests/test_cms.py
+++ b/tests/test_cms.py
@@ -109,7 +109,7 @@ class TestStrictChecker(TestSumCMS):
         overwrite_file(self.task_dir, "check.py", "check_strict.py")
 
 
-class TestDirtySamle(TestSumCMS):
+class TestDirtySample(TestSumCMS):
     """Sample without newline at the end."""
 
     def expecting_success(self):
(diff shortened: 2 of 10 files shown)


pisek/config/config_types.py
```diff
@@ -28,6 +28,11 @@ class JudgeType(StrEnum):
     opendata_v1 = "opendata-v1"
 
 
+class FormatType(StrEnum):
```
5K1PY, line 31: Maybe ``DataFormat`` would be better? ``FormatType`` doesn't say anything about what it describes.

tests/test_cms.py
```diff
@@ -121,6 +121,64 @@ def modify_task(self):
...
+            raw_config["tests"]["checker"] = ""
+
+        modify_config(self.task_dir, modification_fn)
+
+
+class TestNoLFInBinaryInput(TestSumCMS):
+    """Input without newline at the end with in_format=binary."""
+
+    def expecting_success(self):
+        return True
```
5K1PY, line 142: Default implementation for this is same, so no reason to override it, right?

tests/test_cms.py
```diff
@@ -121,6 +121,64 @@ def modify_task(self):
...
+            raw_config["tests"]["checker"] = ""
+
+        modify_config(self.task_dir, modification_fn)
+
+
+class TestNoLFInBinaryOutput(TestSumCMS):
+    """Output without newline at the end with out_format=binary."""
+
+    def expecting_success(self):
+        return True
```
5K1PY, line 171: Same thing here.

5K1PY (changes requested):

tests/test_cms.py
SvizelPritula: This is for consistency with other "change something and run pisek" tests that set it (eg. [here](https://github.com/kasiopea-org/pisek/blob/42098fac2f0c99969dd1550f0bd0939e5d03c483/tests/test_kasiopea.py#L64)).

pisek/config/config_types.py
SvizelPritula: Sure. `DataFormat` is also pretty meaningless, but it would be consistent with the usage of "data" in pisek.

tests/test_cms.py
5K1PY: Looks like we are not very consistent already :sweat_smile:

pisek/config/config_types.py
SvizelPritula: Done.

5K1PY (approved):
:shipit:

PR #178, opened 2024-06-06
Add support for faster whitespace tolerant judge

### LINKED ISSUE #173: Tokenizer
Diff has quadratic complexity. Better replace it with something linear.

SvizelPritula:
This PR imports KSP's `judgelib` judge utility library along with the `judge-token` program. It also adds support for using it instead of `diff`.

Both `out_check=diff` and `out_check=tokens` are supported. If we wan't to deprecate `diff` or alias it to `tokens`, we can easily do so in the future. It also supports all options supported by KSP's opendata tool.

One slight difference between this and KSP's implementation is that the floating point mode is disabled by setting `tokens_float_rel_error` and `tokens_float_abs_error` to `X` instead of the empty string.

Closes #173.

### DIFF
--- a/config-v3-documentation
+++ b/config-v3-documentation
@@ -103,7 +103,7 @@ tokens_float_abs_error=1e-30
 # will be parsed and compared with a given error margin
 # Any tokens that can't be parsed as a float will be compared character-by-character
 # If used, both of these options must be specified
-# Set both options to X to disable
+# To explicitly disable float checking, set both options to the empty string
 
 [test01]
 # Section for each subtask (indexed from one)
--- a/config-v3-documentation
+++ b/config-v3-documentation
@@ -62,6 +62,7 @@ checker=check
 out_check=judge
 # Describes how to check outputs
 #   - diff: checks that contestant output equals correct output (ignores whitespace) (default)
+#   - tokens: uses a fast, versatile file equality checker (ignores whitespace)
 #   - judge: if there can be multiple solutions, checks with program (called 'judge')
 # In communication, only judge is allowed
 
--- a/config-v3-documentation
+++ b/config-v3-documentation
@@ -61,7 +61,7 @@ checker=check
 
 out_check=judge
 # Describes how to check outputs
-#   - diff: checks that contestant output equals correct output (ignores whitespace) (default)
+#   - diff: checks that contestant output equals correct output (ignores whitespace)
 #   - tokens: uses a fast, versatile file equality checker (ignores whitespace)
 #   - judge: if there can be multiple solutions, checks with program (called 'judge')
 # In communication, only judge is allowed
--- a/config-v3-documentation
+++ b/config-v3-documentation
@@ -86,6 +86,25 @@ judge_needs_out=0
 # Set to 1 if judge needs input/output
 # Defaults to 1
 
+tokens_ignore_newlines=0
+# Only for out_check=tokens
+# If set to 1, newline characters will be ignored when checking the output,
+# as if they were any other whitespace characters
+# By default, newline characters are only ignored at the end of the file
+
+tokens_ignore_case=0
+# Only for out_check=tokens
+# If set to 1, ASCII characters will be compared in a case-insensitive manner
+
+tokens_float_rel_error=0.00001
+tokens_float_abs_error=1e-30
+# Only for out_check=tokens
+# When these options are specified, floating-point numbers
+# will be parsed and compared with a given error margin
+# Any tokens that can't be parsed as a float will be compared character-by-character
+# If used, both of these options must be specified
+# Set both options to X to disable
+
 [test01]
 # Section for each subtask (indexed from one)
 # Keys default to [all_tests] keys if test is not set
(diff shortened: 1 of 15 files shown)


config-v3-documentation
```diff
@@ -62,6 +62,7 @@ checker=check
 out_check=judge
 # Describes how to check outputs
 #   - diff: checks that contestant output equals correct output (ignores whitespace) (default)
+#   - tokens: uses a fast, versatile file equality checker (ignores whitespace)
```
5K1PY, line 65: Isn't tokens a better default?

config-v3-documentation
```diff
@@ -85,6 +86,25 @@ judge_needs_out=0
...
+# If set to 1, ASCII characters will be compared in a case-insensitive manner
+
+tokens_float_rel_error=0.00001
+tokens_float_abs_error=1e-30
+# Only for out_check=tokens
+# When these options are specified, floating-point numbers
+# will be parsed and compared with a given error margin
+# Any tokens that can't be parsed as a float will be compared character-by-character
+# If used, both of these options must be specified
+# Set both options to X to disable
```
5K1PY, line 106: I don't really know about the `X`. We use it in different context with a different meaning. I would much prefer either `@` value or empty to disable.

pisek/config/task_config.py
```diff
@@ -226,6 +237,18 @@ def validate_model(self):
                 {"task_type": self.task_type, "out_check": self.out_check},
             )
 
+        if (self.tokens_float_abs_error == None) != (
```
5K1PY, line 240: Nit: Pyhton style prefers ``is None`` instead of ``== None``.

config-v3-documentation
SvizelPritula: Probably. I wanted to keep the default unchanged for backwards compatibility, but that might not be necessary. There aren't any Kasiopea tasks that doesn't set `out_check` explicitly.

pisek/jobs/parts/tools.py
```diff
@@ -38,6 +38,7 @@ def _get_jobs(self) -> list[Job]:
         jobs: list[Job] = [
             PrepareMinibox(self._env),
             PrepareTextPreprocessor(self._env),
+            PrepareJudgeToken(self._env),
```
5K1PY, line 41: Do we really want to always compile this?

pisek/jobs/parts/tools.py
```diff
@@ -100,6 +101,41 @@ def _run(self):
             raise PipelineItemFailure("Text preprocessor compilation failed.")
 
 
+class PrepareJudgeToken(TaskJob):
```
5K1PY, line 104: Nit: I would call it ``PrepareTokenJudge`` to have same name

5K1PY, line 104: Nit: I would call it ``PrepareTokenJudge`` to have same name as

5K1PY, line 104: Nit: I would call it ``PrepareTokenJudge`` to have same name as ``RunTokenJudge``.

5K1PY (changes requested):

config-v3-documentation
5K1PY: I was thinking about adding it to update script

SvizelPritula: What do you want to add to the update script? Adding `out_check=diff` if `out_check` isn't specified?

5K1PY: Yup

pisek/jobs/parts/tools.py
SvizelPritula: I was unsure if I should use the word order used by the program `judge-token` or the more grammatical order. Apparently, I chose both. Fixed.

pisek/jobs/parts/tools.py
SvizelPritula: Fixed. The judge is now compiled on demand.

pisek/config/task_config.py
SvizelPritula: Fixed.

config-v3-documentation
SvizelPritula: I thought it was kind of similar to `max_submissions`. I've changed the unset value to the empty string.

config-v3-documentation
SvizelPritula: Turns out the documentation was wrong: The `out_check` key has no default and is mandatory. I've fixed the documentation.

config-v3-documentation
5K1PY: > I thought it was kind of similar to `max_submissions`. I've changed the unset value to the empty string.

Something like `@unlimited` could be better, but that's up to another PR.

5K1PY: > I thought it was kind of similar to `max_submissions`. I've changed the unset value to the empty string.

Can you also update the documentation?

SvizelPritula: Fixed.

5K1PY (approved):
LGTM. Great work!

PR #179, opened 2024-06-13
Add support for non-sample static tests

SvizelPritula:
Currently, the CMS importer will only search the static input directory for samples only. This PR makes it search for each input glob in both the static directory and the directory with generated inputs.


5K1PY (approved):
LGTM. Would be even nicer, if we didn't overshadow python default functions (``input``).

SvizelPritula:
> Would be even nicer, if we didn't overshadow python default functions (`input`).

Sure, but I don't think we'll ever want to call `input` from this function. :)

PR #185, opened 2024-06-28
CompletenessCheck: Add no redundant lines in CMS judge check

### LINKED ISSUE #184: Check CMS judge stdout & stderr has only one line
Warning, fail on `--strict`.

5K1PY:
Fixes #184

### DIFF
--- a/pisek/jobs/parts/completeness_check.py
+++ b/pisek/jobs/parts/completeness_check.py
@@ -71,7 +71,9 @@ class CompletenessCheck(TaskJobManager):
                 with open(judge_out.path) as f:
                     lines = f.read().rstrip().split("\n")
                 if len(lines) > 1:
-                    return self._warn(f"{judge_out:p} contains multiple lines")
+                    self._warn(f"{judge_out:p} contains multiple lines")
+                    if self._env.verbosity <= 0:
+                        return
 
     def _evaluate(self) -> None:
         self._check_dedicated_solutions()
--- a/pisek/jobs/parts/completeness_check.py
+++ b/pisek/jobs/parts/completeness_check.py
@@ -71,7 +71,7 @@ class CompletenessCheck(TaskJobManager):
             for judge_out in self._get_judge_outs():
                 with open(judge_out.path) as f:
                     lines = f.read().rstrip().split("\n")
-                if len(lines) != 1:
+                if len(lines) > 1 or lines[0] == "":  # Python splitting is weird
                     self._warn(f"{judge_out:p} should contain exactly one line")
                     if self._env.verbosity <= 0:
                         return
--- a/pisek/jobs/parts/completeness_check.py
+++ b/pisek/jobs/parts/completeness_check.py
@@ -10,6 +10,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+from pisek.utils.paths import TaskPath
+from pisek.config.config_types import JudgeType
 from pisek.jobs.jobs import Job
 from pisek.jobs.parts.task_job import TaskJobManager, SOLUTION_MAN_CODE
 
@@ -22,6 +24,14 @@ class CompletenessCheck(TaskJobManager):
 
     def _get_jobs(self) -> list[Job]:
         return []
+
+    def _get_judge_outs(self) -> set[TaskPath]:
+        judge_outs = set()
+        for solution in self._env.solutions:
+            judge_outs |= self.prerequisites_results[f"{SOLUTION_MAN_CODE}{solution}"][
+                "judge_outs"
+            ]
+        return judge_outs
 
     def _check_solution_succeeds_only_on(
         self, sol_name: str, subtasks: list[int]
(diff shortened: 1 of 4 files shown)


pisek/jobs/parts/completeness_check.py
```diff
@@ -52,3 +63,18 @@ def _evaluate(self) -> None:
 
                 if not ok:
                     self._warn(f"{subtask.name} has no dedicated solution")
+
+    def _check_cms_judge(self) -> None:
```
SvizelPritula, line 67: Why is this check part of the completeness check? Doesn't it belong in the judge job?

SvizelPritula (changes requested):

pisek/jobs/parts/completeness_check.py
5K1PY: It's too early in the judge manager, becaause judge outs are not generated yet.

SvizelPritula: Sure, but the exit code of the judge also isn't checked in the completeness check.

5K1PY: Yeah, but you couldn't just print one multilne file that way.

SvizelPritula: I'm not sure what you mean.

5K1PY: Well if verbosity <= 0, we only print one multiline judge out.

5K1PY: > Sure, but this way the check is only ran if all solutions are tested.

Now runs always.

> Also, is the check ran on the results from chaos monkey?

Added.

SvizelPritula: It still feels out of place in the completeness check. This prevents the check from being cached properly, meaning that all judge outputs have to be opened and examined on every change. However, this might be the best we can do without architectural changes.

SvizelPritula: One random idea I had was that each judge job could check it's own output and store the result. The only responsibility of the completeness check part would be to deduplicate and report the warnings.

SvizelPritula:
It might be worth considering making this an error instead of a warning. Sure, the CMS docs state that extra lines are ignored, but extra lines are almost always a programmer error.

pisek/jobs/parts/completeness_check.py
5K1PY: > It still feels out of place in the completeness check.

I agree. But I don't really have good solution yet.

5K1PY: > One random idea I had was that each judge job could check it's own output and store the result. The only responsibility of the completeness check part would be to deduplicate and report the warnings.

But that is not much different than the current version, is it? Plus having this thing cached is going to  save only little time compared to whole run.

5K1PY:
> It might be worth considering making this an error instead of a warning. Sure, the CMS docs state that extra lines are ignored, but extra lines are almost always a programmer error.

Or a helper for solution debuging... And that has lot of sense to enforce only in final version of the task (`--strict`).

pisek/jobs/parts/completeness_check.py
SvizelPritula: I don't actually know how long opening and reading so many files will take, it might very well be fine. I just thought it might make sense to read the verdict files only once.

5K1PY: > I don't actually know how long opening and reading so many files will take, it might very well be fine.

Tested this on bigger task (2000+ files) and it's basically instant.

> I just thought it might make sense to read the verdict files only once.

We do open solution outputs multiple times btw :innocent:

pisek/jobs/parts/completeness_check.py
```diff
@@ -52,3 +64,20 @@ def _evaluate(self) -> None:
...
+        """Checks that cms judge's stdout & stderr contains only one line."""
+        if self._env.config.judge_type == JudgeType.cms:
+            for judge_out in self._get_judge_outs():
+                with open(judge_out.path) as f:
+                    lines = f.read().rstrip().split("\n")
+                if len(lines) > 1:
```
SvizelPritula, line 74: Actually, maybe we should check there is exactly one line. The judge job obviously checks this is the case for stdout, but it doesn't even open stderr.

5K1PY: Good point.

pisek/jobs/parts/completeness_check.py
```diff
@@ -71,8 +71,8 @@ def _check_cms_judge(self) -> None:
             for judge_out in self._get_judge_outs():
                 with open(judge_out.path) as f:
                     lines = f.read().rstrip().split("\n")
```
SvizelPritula, line 73: Won't `split` with `sep` specified always return a non-empty list?

5K1PY: First!

Yeah this behaviour is really weird...

SvizelPritula: > Yeah this behaviour is really weird...

Not really...

SvizelPritula (approved):
LGTM

pisek/jobs/parts/completeness_check.py
SvizelPritula: Nit: It might be nice to have two separate messages for > 1 and < 1, but you'll probably want to look at the file anyway.

PR #186, opened 2024-07-02
ConfigDocs: Document in/out_format

### LINKED ISSUE #176: Document in_format & out_format
Add them to ``config-v3-documentation.``

5K1PY:
Fixes #176

### DIFF
--- a/config-v3-documentation
+++ b/config-v3-documentation
@@ -65,7 +65,7 @@ out_format=text
 #   text    ASCII plain text (LF newlines)
 #   binary  can be anything
 #
-# TODO: Server mode
+# TODO: Mention auto normalization in Server mode when it's done (#137)
 
 checker=check
 # Program (without suffix) that validates inputs
--- a/config-v3-documentation
+++ b/config-v3-documentation
@@ -55,6 +55,18 @@ data_subdir=tests/
 in_gen=src/gen
 # Source code of generator (without suffix)
 
+in_format=text
+# Format of input:
+#   text    ASCII plain text (LF newlines)
+#   binary  can be anything
+
+out_format=text
+# Format of output:
+#   text    ASCII plain text (LF newlines)
+#   binary  can be anything
+#
+# TODO: Server mode
+
 checker=check
 # Program (without suffix) that validates inputs
 # No value means no checking


config-v3-documentation
```diff
@@ -55,6 +55,20 @@ data_subdir=tests/
...
+#           subject to normalization (newlines, encoding)
+#   binary  application/binary
+
+out_format=text
+# Format of output:
+#   text    text/plain in ASCII (default)
+#           subject to normalization (newlines, encoding)
+#   binary  application/binary
+#
+# TODO: Server mode
```
SvizelPritula, line 70: I'm not sure I understand the point of this TODO. Will  the format or semantics of this option check anything when the server mode is implemented?

config-v3-documentation
```diff
@@ -55,6 +55,20 @@ data_subdir=tests/
 in_gen=src/gen
 # Source code of generator (without suffix)
 
+in_format=text
+# Format of input:
+#   text    text/plain in ASCII (default)
+#           subject to normalization (newlines, encoding)
+#   binary  application/binary
```
SvizelPritula, line 62: This is not a real MIME type. Did you mean `application/octet-stream`?

config-v3-documentation
```diff
@@ -55,6 +55,20 @@ data_subdir=tests/
 in_gen=src/gen
 # Source code of generator (without suffix)
 
+in_format=text
+# Format of input:
+#   text    text/plain in ASCII (default)
```
SvizelPritula, line 60: The `text/plain` MIME type requires that lines end with CRLF, but písek enforces LF.

SvizelPritula (changes requested):

config-v3-documentation
5K1PY: Yup, that's why is there:
> subject to normalization (newlines, encoding)

config-v3-documentation
5K1PY: Right. But I would much prefer having something imediately clear...

config-v3-documentation
5K1PY: Improved the comment.

config-v3-documentation
SvizelPritula: If it's not supposed to be a MIME type, then I don't understand what you mean by "application/binary". Do you mean the input/output will be either an application or a sequence of bits?

config-v3-documentation
SvizelPritula: Sure, but písek doesn't normalize the files to `text/plain`. If we want our input files to be `text/plain` files, písek should replace all LFs by CRLFs before giving them as inputs to solutions or exporting them.

SvizelPritula (approved):
The new version makes way more sense, thanks!

PR #188, opened 2024-07-03
Rework CLI

5K1PY:
- [x] Test only on test
- [x] Fix flags and remove once called functions
- [x] Fix tests
- [x] Check CMS didn't broke
- [x] Add changes to README


SvizelPritula (approved):
LGTM.

It's sad that we have to make the main way to run pisek more complicated, but I don't see a way around it without complicated hacks. (Maybe we could create two parsers and try both, or something like that.)

PR #189, opened 2024-07-03
Update readme

5K1PY:


README.md
```diff
@@ -33,11 +36,11 @@ It tests all task parts (generator, checker, solutions and judge).
 ### Task testing overview
 
 What pisek verifies (roughly in order):
- - Samples exist and are not empty
+ - Samples exist
  - The generator generates inputs
     - In Kasiopea mode the generator respects the seed and is deterministic
  - The checker accepts all inputs
-    - It rejects inputs for harder subtasks
+    - If `checker_distinguishes_subtasks`, it rejects inputs for harder subtasks
```
SvizelPritula, line 43: It may be clearer if we specify what the value of `checker_distinguishes_subtasks` should be:

```suggestion
    - If `checker_distinguishes_subtasks=on`, it rejects inputs for harder subtasks
```

README.md
```diff
@@ -49,6 +52,10 @@ What pisek verifies (roughly in order):
     - They don't contain unprintable characters
     - They have a newline at the end
     - In Kasiopea mode the files are reasonably small
+ - Task is complete
+    - In CMS there are no redundant lines in judge's stdout/stderr.
+    - If `solution_for_each_subtask`, there exists dedicated solution for each subtask
```
SvizelPritula, line 57: You should use articles when writing English text.

```suggestion
 - The task is complete
    - In CMS there are no redundant lines in the judge's stdout/stderr.
    - If `solution_for_each_subtask`, there exists a dedicated solution for each subtask
```

README.md
```diff
@@ -49,6 +52,10 @@ What pisek verifies (roughly in order):
     - They don't contain unprintable characters
     - They have a newline at the end
     - In Kasiopea mode the files are reasonably small
+ - Task is complete
+    - In CMS there are no redundant lines in judge's stdout/stderr.
```
SvizelPritula, line 56: This is the only sentence in this list that ends with a period.

```suggestion
    - In CMS there are no redundant lines in judge's stdout/stderr
```

SvizelPritula (changes requested):

README.md
```diff
@@ -49,6 +52,10 @@ What pisek verifies (roughly in order):
     - They don't contain unprintable characters
     - They have a newline at the end
     - In Kasiopea mode the files are reasonably small
+ - The task is complete
+    - In CMS there are no redundant lines in the judge's stdout/stderr
```
SvizelPritula, line 56: I noticed one more issue. For consistency with lines 41 and 54, we should use this:

```suggestion
    - In CMS mode there are no redundant lines in the judge's stdout/stderr
```

SvizelPritula (changes requested):

README.md
5K1PY: This should be better. Mostly because CMS mode is an illusion :D

SvizelPritula: Kasiopea mode is also mentioned on line 41.

README.md
```diff
@@ -48,7 +51,11 @@ What pisek verifies (roughly in order):
     - They are in the correct encoding
     - They don't contain unprintable characters
     - They have a newline at the end
-    - In Kasiopea mode the files are reasonably small
+    - The files are reasonably small
+ - The task is complete
+    - CMS judge has no redundant lines in its stdout/stderr
```
SvizelPritula, line 56: There needs to be an article here...

```suggestion
    - A CMS judge has no redundant lines in its stdout/stderr
```

README.md
```diff
@@ -33,11 +36,11 @@ It tests all task parts (generator, checker, solutions and judge).
 ### Task testing overview
 
 What pisek verifies (roughly in order):
- - Samples exist and are not empty
+ - Samples exist
  - The generator generates inputs
-    - In Kasiopea mode the generator respects the seed and is deterministic
+    - Kasiopea generator respects the seed and is deterministic
```
SvizelPritula, line 41: There needs to be an article here...


```suggestion
    - A Kasiopea generator respects the seed and is deterministic
```

SvizelPritula (changes requested):

SvizelPritula (approved):
LGTM

PR #192, opened 2024-07-16
Folder reform: pisek/jobs/parts -> pisek/task_jobs

5K1PY:
Bit of folder reform.

As we use `jobs/` as a library, it doesn't make to have `jobs/parts/` inside the library. Also allows some more nesting without files being too deep.


SvizelPritula (approved):
LGTM

Issue #108, opened 2023-10-17
Tests for server mode


5K1PY:
Currently we have server implementation in [pisek/server.py](/kasiopea-org/pisek/blob/master/pisek/server.py). We could use some test for it.

Issue #168, opened 2024-05-09
New generator format


5K1PY:
Add new generator format with the following convention:
```
./gen
```
Prints all inputs generator can generate.
```
./gen [input_name] [seed]
```
Generates input on given seed.

PR #197, opened 2024-08-02
Generate inputs on the fly

### LINKED ISSUE #137: Rewrite server mode
And #108

### LINKED ISSUE #170: Remove contest_type
Blocked on #168

5K1PY:
Rewrote generator interface for easier implementation of various generator types.

Generating inputs on the fly.

Replaced `constest_type` with `gen_type` (Fixes #170)

Removed server mode (Fixes #137, #108 kind of)


pisek/config/task_config.py
```diff
@@ -369,9 +356,9 @@ def validate_globs(cls, value: list[str], info: ValidationInfo) -> list[str]:
         for glob in value:
             if glob == "@ith":
                 glob = f"{info.data['num']:02}*.in"
-            if not glob.endswith(".in"):
+            if not glob.endswith("*.in"):
```
SvizelPritula, line 359: This is a pretty significant change, which will break old tasks. Is it really needed?

pisek/jobs/jobs.py
```diff
@@ -83,7 +83,9 @@ def __init__(self, name: str) -> None:
         self.dirty = False  # Prints something to console?
 
         self.prerequisites = 0
-        self.required_by: list[tuple["PipelineItem", Optional[str]]] = []
+        self.required_by: list[
+            tuple["PipelineItem", Optional[str], Callable[[Any], bool]]
+        ] = []
```
SvizelPritula, line 88: Nit: It might be beneficial to use a named tuple, since it isn't obvious what each element means.

pisek/jobs/task_pipeline.py
```diff
@@ -47,53 +45,54 @@ def __init__(self, env: Env):
...
-                        f"{SOLUTION_MAN_CODE}{env.config.primary_solution}",
-                    )
-                )
-                solutions.append(primary_solution)
+            first_solution_name = (
+                env.config.primary_solution
+                if env.config.judge_needs_out
+                or env.config.primary_solution in env.solutions
+                else env.solutions[0]
+            )
```
SvizelPritula, line 70: Why does the first solution need to be always handled differently?

pisek/task_jobs/data/data.py
```diff
@@ -0,0 +1,102 @@
...
+        super().__init__(
+            env=env,
+            name=name,
+            **kwargs,
+        )
+        self.data = data
+
+
+class LinkData(DataJob):
+    """Copy data to into dest folder."""
```
SvizelPritula, line 34: Does it copy or link?

pisek/task_jobs/data/manager.py
```diff
@@ -0,0 +1,122 @@
...
+        all_input_infos: list[InputInfo] = [
+            InputInfo.static(inp.name.removesuffix(".in")) for inp in static_inputs
+        ] + self.prerequisites_results[GENERATOR_MAN_CODE]["inputs"]
+        all_input_infos.sort(key=lambda info: info.name)
+
+        self._input_infos: dict[int, list[InputInfo]] = {}
+        for num, sub in self._env.config.subtasks.items():
+            self._input_infos[sub.num] = []
+            for input_info in all_input_infos:
+                if sub.in_subtask(input_info.task_path(self._env, 25265).name):
```
SvizelPritula, line 62: 25265 should probably be a constant.

pisek/task_jobs/data/manager.py
```diff
@@ -0,0 +1,122 @@
...
+                jobs.append(IsClean(self._env, path))
+            if self._env.config.limits.input_max_size != 0:
+                jobs.append(InputSmall(self._env, path))
+
+        for path in static_outputs:
+            jobs.append(LinkOutput(self._env, path))
+            if self._env.config.out_format == DataFormat.text:
+                jobs.append(IsClean(self._env, path))
+            if self._env.config.limits.output_max_size != 0:
+                jobs.append(OutputSmall(self._env, path))
```
SvizelPritula, line 99: Why do static inputs need this much special handling?

pisek/task_jobs/data/manager.py
```diff
@@ -0,0 +1,122 @@
...
+        used_inputs = set(sum(self._input_infos.values(), start=[]))
+        unused_inputs = list(set(all_input_infos) - used_inputs)
+        if len(unused_inputs):
+            if self._env.verbosity <= 0:
+                CUTOFF = 3
+                unused_inputs_text = ", ".join(
+                    map(lambda inp: inp.name, unused_inputs[:CUTOFF])
+                )
+                if len(unused_inputs) > CUTOFF:
+                    unused_inputs_text += ",..."
```
SvizelPritula, line 78: This could be changed to a `…` character.

pisek/task_jobs/data/manager.py
```diff
@@ -0,0 +1,122 @@
...
+    def __init__(self) -> None:
+        super().__init__("Processing data")
+
+    def _get_jobs(self) -> list[Job]:
+        static_inputs = self.globs_to_files(
+            self._env.config.input_globs, TaskPath.static_path(self._env, ".")
+        )
+        static_outputs = self.globs_to_files(
+            map(
+                lambda g: re.sub(r"\.in$", ".out", g),
```
SvizelPritula, line 39: Why doesn't this use `replace_suffix`?

pisek/task_jobs/generator/cms_old.py
```diff
@@ -0,0 +1,78 @@
...
+class CmsOldGenerate(GenerateInput):
+    def __init__(
+        self, env: Env, generator: TaskPath, input_info: InputInfo, seed: int, **kwargs
+    ) -> None:
+        super().__init__(
+            env,
+            generator,
+            input_info,
+            seed,
+            name=f"Serve {input_info.task_path(env).name}",
```
SvizelPritula, line 69: I don't think this name tells you much.

pisek/task_jobs/generator/manager.py
```diff
@@ -0,0 +1,220 @@
...
+    OpendataV1ListInputs,
+    OpendataV1Generate,
+    OpendataV1TestDeterminism,
+)
+
+SEED_RANGE = range(0, 16**8)
+
+
+class PrepareGenerator(TaskJobManager):
+    """Manager that generates inputs and test generator."""
```
SvizelPritula, line 48: The manager doesn't generate a test generator. Did you mean "test**s** the generator"?

pisek/task_jobs/generator/manager.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _input_info_jobs(self, input_info: InputInfo, subtask: int) -> list[Job]:
+        repeat = input_info.repeat * (
+            self._env.repeat_inputs if input_info.seeded else 1
+        )
+
+        jobs: list[Job] = []
+
+        random.seed(4)  # Reproducibility!
+        seeds = random.sample(SEED_RANGE, repeat)
```
SvizelPritula, line 113: If there is any change of implementing parallelism in the future, it might be a good idea to construct a `Random` instance instead of changing global state.

pisek/task_jobs/generator/manager.py
```diff
@@ -0,0 +1,220 @@
...
+
+        jobs: list[Job] = []
+        if len(seeds) == 1:
+            while (seed := random.choice(SEED_RANGE)) == seeds[0]:
+                pass
+            seeds.append(seed)
+            jobs += [self._generate_input_job(input_info, seed)]
+
+        jobs.append(
+            check_seeded := GeneratorRespectsSeed(self._env, input_info, *seeds[:2])
```
SvizelPritula, line 201: Nit: it might be useful to check all pairs, but that can wait.

pisek/task_jobs/solution.py
```diff
@@ -55,32 +59,71 @@ def _get_jobs(self) -> list[Job]:
...
+        if not self._is_first:
+            return []
+        return super()._generate_input_jobs(input_info, seed, subtask, test_determinism)
+
+    def _respects_seed_jobs(
+        self, input_info: InputInfo, seeds: list[int], subtask: int
+    ) -> list[Job]:
+        if not self._is_first:
+            return []
+        return super()._respects_seed_jobs(input_info, seeds, subtask)
```
SvizelPritula, line 88: I don't understand what those methods are for.

pisek/task_jobs/solution.py
```diff
@@ -55,32 +59,71 @@ def _get_jobs(self) -> list[Job]:
...
+                jobs.append(out_clean := IsClean(self._env, run_batch_sol.output))
+                out_clean.add_prerequisite(
+                    run_batch_sol,
+                    condition=lambda r: r.kind == RunResultKind.OK,
+                )
+            if self._env.config.limits.output_max_size != 0:
+                jobs.append(out_small := OutputSmall(self._env, run_batch_sol.output))
+                out_small.add_prerequisite(run_judge)
+
+        elif self._env.config.task_type == "communication":
```
SvizelPritula, line 114: Shouldn't this be `TaskType.communication`?

pisek/task_jobs/solution.py
```diff
@@ -90,7 +133,7 @@ def _create_batch_jobs(
         )
         run_solution.add_prerequisite(self._compile_job)
 
-        if sub_num == 0:
+        if subtask == 0:
```
SvizelPritula, line 136: Why is subtask 0 treated differently here? Shouldn't the important thing be if it's static or not?

pisek/task_jobs/data/manager.py
```diff
@@ -0,0 +1,122 @@
...
+                self._env.config.input_globs,
+            ),
+            TaskPath.static_path(self._env, "."),
+        )
+
+        if self._env.config.task_type != "communication":
```
SvizelPritula, line 45: Shouldn't this be `TaskType.communication`?

pisek/task_jobs/judge.py
```diff
@@ -56,7 +56,7 @@ def _get_jobs(self) -> list[Job]:
         elif self._env.config.out_check == OutCheck.tokens:
             jobs.append(comp := PrepareTokenJudge(self._env))
 
-        samples = self._get_samples()
+        samples = self._get_static_samples()
         if self._env.config.task_type == "communication":
```
SvizelPritula, line 60: Shouldn't this be `TaskType.communication`?

pisek/task_jobs/solution.py
```diff
@@ -55,32 +59,71 @@ def _get_jobs(self) -> list[Job]:
...
+    def _solution_jobs(
+        self, input_info: InputInfo, seed: int, subtask: int
+    ) -> list[Job]:
+        input_path = input_info.task_path(self._env, seed)
+
+        jobs: list[Job] = []
+
+        run_sol: RunSolution
+        run_judge: RunJudge
+        if self._env.config.task_type == "batch":
```
SvizelPritula, line 99: Shouldn't this be `TaskType.batch`?

SvizelPritula (changes requested):
There are some issues.

pisek/__main__.py
```diff
@@ -158,7 +158,7 @@ def add_argument_dataset(parser):
         "--inputs",
         "-n",
         type=int,
-        default=5,
+        default=1,
```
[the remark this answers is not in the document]
5K1PY: Mostly as a preperation for #168. While repeating 5 times is much more safer, if you write in your generator ``repeat=2`` it would be quite surprising to generate it 10 times.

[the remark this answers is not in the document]
SvizelPritula: Sure, but you'll unlikely to use `repeat` at all for Kasiopea or KSP-style tasks.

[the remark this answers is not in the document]
5K1PY: Sure, but we intend to have this generator for CMS style tasks.

pisek/task_jobs/generator/manager.py
5K1PY: The problem with checking all pairs is that you can get a collision. (Like if input is a power of 2 lesser than 1024...)

pisek/task_jobs/solution.py
5K1PY: So we don't generate the jobs again if we are not on first solution.

Jobs handling still requires some not trival running time.

pisek/jobs/task_pipeline.py
5K1PY: It generates the inputs, so other solutions depend on it.

pisek/task_jobs/data/manager.py
5K1PY: I thought it would be better if we check them before testing Judge with them.

pisek/task_jobs/generator/manager.py
SvizelPritula: That can also happen if you only check two...
Maybe the correct test is if all are equal.

pisek/task_jobs/generator/cms_old.py
5K1PY: I know. But does it really matter as you wont really see this? Or you have some better suggestion?

pisek/jobs/task_pipeline.py
SvizelPritula: Wouldn't it be easier for `TaskPipeline` to create the generation job instead of having `TaskPipeline` pick a random different job to task with creating them?

5K1PY: You mean adding ``RunGenerator()`` there?

pisek/task_jobs/generator/cms_old.py
SvizelPritula: Copy {input_info.task_path(env).name}?

pisek/jobs/task_pipeline.py
SvizelPritula: Yes

pisek/__main__.py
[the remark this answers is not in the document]
SvizelPritula: But we still intend to use Pisek for Kasiopea, right?

pisek/task_jobs/generator/cms_old.py
5K1PY: And does that say anything more useful?

pisek/__main__.py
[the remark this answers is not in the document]
5K1PY: Sure but you can just add `-n5` if you want?

pisek/task_jobs/generator/cms_old.py
SvizelPritula: Which meaning of "serve" do you use in the current name?

pisek/__main__.py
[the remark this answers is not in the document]
SvizelPritula: I think 5 was a useful default.

pisek/task_jobs/generator/cms_old.py
5K1PY: **verb** as in aid, help; supply

pisek/__main__.py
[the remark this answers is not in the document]
5K1PY: I also agree that 5 is a useful default for opendata tasks. But I think it's terrible default for CMS.

pisek/jobs/task_pipeline.py
5K1PY: This way we can run solution straight away which means you do not have to wait for generator to generate all inputs.

pisek/task_jobs/generator/manager.py
5K1PY: I would leave paralelism for the far far future and not bother with it now.

pisek/config/task_config.py
5K1PY: You're right. Best to leave this as it.

pisek/__main__.py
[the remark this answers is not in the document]
SvizelPritula: Maybe we could load the default from the config file, with the expectation that it'll always be inherited from the task type config. Alternatively, we could change it based on `fail_mode`.

pisek/jobs/task_pipeline.py
SvizelPritula: Why does that require that the generator jobs are created by a solution?

pisek/task_jobs/generator/cms_old.py
SvizelPritula: I've never heard "serve" used in this way, I'd associate "serve" with something that's done repeatedly, on demand.

pisek/task_jobs/generator/manager.py
SvizelPritula: I'd still prefer to avoid unneeded global state,  especially because it's a simple change (`random.Random(4).sample(...)`).

pisek/task_jobs/generator/cms_old.py
5K1PY: Well you can check it out on [Thesaurus](https://www.thesaurus.com/browse/serve)

pisek/jobs/task_pipeline.py
5K1PY: We want to run jobs this way:
- Generate input
- Check input
- Run solution
- Judge output
- Check output

We can change this manager to have a different name and display message, but it's still going to be manager of the first solution at the end of the day.

SvizelPritula: If were running jobs in creation order, sure.
It might be cleaner to have each solution job try to find the correct generator jobs and create them if they don't exist, but that could be a lot of work.

Why is creating jobs so resource intensive anyway?

pisek/task_jobs/data/manager.py
SvizelPritula: Could we at least deduplicate the code to create all the check job creation with `InputInfoMixin._generate_input_jobs` and `SolutionManager._solution_jobs`?

pisek/task_jobs/generator/cms_old.py
SvizelPritula: Thesaurus says that "deliver" is a synonym for "serve", but "serve" isn't a synonym for "deliver".
None of the [dictionary definitions](https://www.dictionary.com/browse/serve#american-serve-verb-used-with-object) seem to match our case.

SvizelPritula (approved):
LGTM

PR #213, opened 2024-08-12
PisekV1: disallow input repetition

### LINKED ISSUE #212: Pisek-v1 generator: Don't allow input being listed multiple times

Tonyl314:
Fixes #212

### DIFF
--- a/pisek/task_jobs/generator/pisek_v1.py
+++ b/pisek/task_jobs/generator/pisek_v1.py
@@ -32,9 +32,14 @@ class PisekV1ListInputs(GeneratorListInputs):
 
     def _run(self) -> list[InputInfo]:
         input_infos = []
+        input_names: set[str] = set()
         self._create_inputs_list()
         for i, line in enumerate(self._get_input_lines()):
-            input_infos.append(self._get_input_info_from_line(line, i))
+            input_info = self._get_input_info_from_line(line, i)
+            if input_info.name in input_names:
+                self._line_invalid(i, line, "Input already listed")
+            input_names.add(input_info.name)
+            input_infos.append(input_info)
         return input_infos
 
     def _get_input_info_from_line(self, line: str, line_number: int) -> InputInfo:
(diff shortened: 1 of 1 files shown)


pisek/task_jobs/generator/pisek_v1.py
```diff
@@ -32,9 +32,14 @@ def __init__(self, env: Env, generator: TaskPath, **kwargs) -> None:
 
     def _run(self) -> list[InputInfo]:
         input_infos = []
+        input_names: set[str] = set()
         self._create_inputs_list()
         for i, line in enumerate(self._get_input_lines()):
-            input_infos.append(self._get_input_info_from_line(line, i))
+            input_info = self._get_input_info_from_line(line, i)
```
5K1PY, line 38: We should `rstrip("\n")` the line here, so we don't get different number of newlines depending on error.

5K1PY (changes requested):

pisek/task_jobs/generator/pisek_v1.py
Tonyl314: Good point. I have added the _rstrip_ to __line_invalid_ because it is lower level.

pisek/task_jobs/generator/pisek_v1.py
```diff
@@ -32,58 +32,64 @@ def __init__(self, env: Env, generator: TaskPath, **kwargs) -> None:
...
+                line_index, line, "For unseeded input 'repeat' must be '1'"
             )
 
         return InputInfo.generated(input_name, **info_args)
 
     def _line_invalid(self, line_index: int, contents: str, reason: str) -> NoReturn:
+        contents = contents.rstrip("\n")
         message = (
-            f"Inputs list invalid (line {line_index}) - {reason}:\n"
+            f"Inputs list invalid (line number {line_index+1}) - {reason}:\n"
```
5K1PY, line 92: Nit: number is a bit redundant here
```suggestion
            f"Inputs list invalid (line {line_index+1}) - {reason}:\n"
```

5K1PY (approved):
:shipit:

pisek/task_jobs/generator/pisek_v1.py
Tonyl314: I think it makes it a bit clearer that it is not an index starting at 0?

5K1PY: I haven't really seen anywhere lines numbered from 0, but it's a detail.

Issue #193, opened 2024-07-31
Implement opendata-v2 judge


5K1PY:
Opendata judge returning 42/43 and allowing metadata.

PR #218, opened 2024-09-09
Support absolute points

### LINKED ISSUE #217: Improve `pisek_log.json` format
For example why solution can't be named source.

5K1PY:
- Allows progress on #193
- Adds adjustable `score_precision`.
- Fixes #217


pisek/visualize.py
```diff
@@ -135,18 +155,23 @@ def _evaluate_results(
...
     ) -> "SolutionResults":
-        if name not in testing_log:
+        if name not in testing_log["solutions"]:
             raise MissingSolution(name)
 
         results = []
-        for result in testing_log[name]["results"]:
+        for test_name, result in testing_log["solutions"][name]["results"].items():
+
+            def tryDecimal(s: Optional[str]) -> Optional[Decimal]:
```
SvizelPritula, line 164: This method should have a `snake_case` name.
I also don't think `try` is the proper word here, something like `optional` or `maybe` would be more appropriate.

pisek/task_jobs/testing_log.py
```diff
@@ -48,23 +52,30 @@ def _evaluate(self) -> None:
...
+                    "relative_points": (
+                        str(sol_res.relative_points)
+                        if isinstance(sol_res, RelativeSolutionResult)
+                        else None
+                    ),
+                    "absolute_points": (
+                        str(sol_res.absolute_points)
+                        if isinstance(sol_res, AbsoluteSolutionResult)
+                        else None
+                    ),
```
SvizelPritula, line 76: I think it would be better to omit the other key entirely instead of setting it to `null`.

SvizelPritula (changes requested):

SvizelPritula (approved):
LGTM

PR #225, opened 2024-09-19
Env: Speed improvements

5K1PY:
Instead of clonning env objects, we just save and clear accesses at right times.

We get up to 2x speed improvement on fully cached tasks :tada:


SvizelPritula (approved):
LGTM

PR #228, opened 2024-09-25
Update config

### LINKED ISSUE #162: There is no way to update the config file
It used to be possible to use `pisek update` to update the config file to version 2. Pisek now does this automatically on startup, but this is in memory only and doesn't actually get written to the config file.

5K1PY:
Fixes #162

### DIFF
--- a/pisek/config/config_tools.py
+++ b/pisek/config/config_tools.py
@@ -1,0 +1,26 @@
+import configparser
+import os
+from typing import Optional
+
+from pisek.config.update_config import update_config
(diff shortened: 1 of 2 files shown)


pisek/config/config_tools.py
SvizelPritula, line 1: Is it necessary to read the config twice?

5K1PY: If we want to check if the config is even valid, yes. The first reading checks things like configs upper in the hierarchy.

SvizelPritula (approved):
LGTM

PR #237, opened 2024-09-27
CMS: Get testcase names from generator

SvizelPritula:
The CMS module currently searches through the `data/generated/` directory to find all the testcases, but that directory might not exist for some generators. With this change, the CMS module will reuse the list generated by the pipeline.


5K1PY (changes requested):
It might be actually nicer to return `TaskPath` objects from the pipeline if you will convert it back anyway.

5K1PY:
But with `TaskPath`s, we also give the full path to the input. This removes the need to guess the folder, to which the input belongs. If we needed to change the name of the inputs folder with current interface, we would break the things that use it.

Overall `TaskPath`s are more resilient to change.

SvizelPritula:
> This removes the need to guess the folder

Doesn't `TaskPath.input_path` handle that? We need to rely on `TaskPath.output_file` anyway.

5K1PY (approved):
Thanks!

> Doesn't TaskPath.input_path handle that? We need to rely on TaskPath.output_file anyway.

There is a certain story about `cms-old` generators having `tests/` hardcoded instead of the first argument...

PR #250, opened 2024-10-06
CMS: Add flag -a to act on active dataset

### LINKED ISSUE #182: CMS datasets
Could be to set dataset as `-d @last` or `-d @active`

SvizelPritula:
This PR adds a flag named `--active-dataset` (or just `-a`) that allows the user to select the active dataset for `cms check` or `cms testing-log` without having to remember its name. 

Somewhat fixes #182


5K1PY (approved):
LGTM

PR #251, opened 2024-10-13
Implement mixed inputs

5K1PY:
Add support for mixed inputs, where input is static and output is generated.


pisek/task_jobs/data/manager.py
```diff
@@ -32,45 +32,65 @@ def __init__(self) -> None:
...
-        all_input_infos: list[InputInfo] = [
-            InputInfo.static(inp.name.removesuffix(".in")) for inp in all_static_inputs
+        all_testcase_infos: list[TestcaseInfo] = [
+            (
+                TestcaseInfo.static(inp.name.removesuffix(".in"))
+                if self._env.config.task_type == TaskType.communication
+                or inp.replace_suffix(".out").exists()
+                else TestcaseInfo.mixed(inp.name.removesuffix(".in"))
+            )
+            for inp in all_static_inputs
```
SvizelPritula, line 63: This logic is partially duplicated from the loop above. The two loops could be merged.

pisek/task_jobs/data/manager.py
```diff
@@ -32,45 +32,65 @@ def __init__(self) -> None:
...
+                inp_path = testcase_info.input_path(self._env, TEST_SEED).name
+                if sub.in_subtask(inp_path):
+                    if (
+                        num == 0
+                        and testcase_info.generation_mode
+                        != TestcaseGenerationMode.static
+                    ):
+                        raise PipelineItemFailure(
+                            f"Sample inputs must be static, but '{inp_path}' is {testcase_info.generation_mode}."
+                        )
```
SvizelPritula, line 82: This seems out of place in this loop.

pisek/jobs/cache.py
```diff
@@ -27,6 +27,11 @@
 SAVED_LAST_SIGNATURES = 5
 
 
+class NoAliasDumper(yaml.Dumper):
+    def ignore_aliases(self, data):
+        return True
```
SvizelPritula, line 32: Why do we need this?

SvizelPritula (changes requested):

pisek/jobs/cache.py
5K1PY: Because we export `CacheEntries` one by one. This will always generate alias with id 1. And ids have to be unique.

SvizelPritula (approved):

PR #252, opened 2024-10-24
Main: Make help descriptions more consistent + add autocomplete

5K1PY:


pisek/__main__.py
```diff
@@ -108,215 +108,215 @@ def add_argument_dataset(parser):
...
     config_subparsers.add_parser(
-        "update", help="Update config to newest version. (Replaces the config.)"
+        "update", help="update config to newest version (replaces the config)"
     )
 
     # ------------------------------- pisek visualize -------------------------------
 
     parser_visualize = subparsers.add_parser(
-        "visualize", help="Show solutions statistics and closeness to limit."
+        "visualize", help="show solutions statistics and closeness to limit"
```
SvizelPritula, line 225: This should probably use the singular form of `solution`.

```suggestion
        "visualize", help="show solution statistics and closeness to limit"
```

pisek/__main__.py
```diff
@@ -82,7 +82,7 @@ def add_argument_description(parser):
             "--description",
             "-d",
             type=str,
-            help="Created the dataset with a description of DESCRIPTION.",
+            help="create the dataset with a description DESCRIPTION",
```
SvizelPritula, line 85: This should be either `a description of DESCRIPTION` or `the description DESCRIPTION`.

SvizelPritula (changes requested):
I found two small issues.

We're somewhat inconsistent in how many articles we use in the help messages, but I don't think that's too big of an issue.

pisek/__main__.py
5K1PY: What? It visualizes multiple solutions.

SvizelPritula: Train tracks also aren't called trains tracks, even though multiple trains run on them.

5K1PY: Right, I'm stupid. Should read next time.

Issue #256, opened 2024-11-07
Handle missing testing log


Tonyl314:
Running `pisek visualize` without a testing log prints an uncaught error. It would be nicer to instead explain the need to first test with the `--testing-log` flag.

5K1PY:
Fixed by #257

PR #257, opened 2024-11-08
Handle missing testing log

Tonyl314:
I catch the error when trying to visualize without a testing log, explain the need to test with `--testing-log`.
Addresses #256.


5K1PY (approved):
LGTM :shipit:

PR #259, opened 2024-11-09
Config: Remove data_subdir

5K1PY:
Remove support for configurable name of `data_subdir` as it is not really necesarry.


SvizelPritula (commented):
I don't think there remains any infrastructure that looks inside of the data directory, so this should be fine.

However, there are some older tasks that set the key and will therefore break.

SvizelPritula (approved):

PR #261, opened 2024-11-16
TestcaseManager: Vary seeds between testcases

### LINKED ISSUE #253: Vary seeds between input types
Currently, when using the new `pisek-v1` judge, all seeded input types will be generated using the same seed / sequence of seeds. This means that different inputs can become correlated, unless the generator mixes the seed and input name together.

SvizelPritula:
Fixes #253.


5K1PY (approved):
LGTM

PR #265, opened 2024-11-16
TaskJob: Don't pass files to CMS judge it didn't request

### LINKED ISSUE #238: Make cms-batch-judge respect `judge_needs`
We could give it files it can't read. (There is `/dev/full` for writing.)

Or some file with too strict permissions.

SvizelPritula:
Fixes #238.


5K1PY (approved):
LGTM

PR #266, opened 2024-11-16
Cache: Fix loading almost empty cache

5K1PY:


pisek/jobs/cache.py
```diff
@@ -142,6 +142,9 @@ def _load(self, env: Env) -> None:
                 return self._new_cache_file()
 
             entries = yaml.full_load(f)
+            if entries is None:
+                return
```
SvizelPritula, line 146: Shouldn't this be something like `return self._new_cache_file()`?

SvizelPritula (approved):
LGTM

pisek/jobs/cache.py
5K1PY: It doesn't really matter

PR #268, opened 2024-11-16
Clean non relevant files

### LINKED ISSUE #229: Clean non-relevant files after testing
Remove files there were not generated in this run of `pisek test`.

5K1PY:
Fixes #229

### DIFF
--- a/pisek/jobs/jobs.py
+++ b/pisek/jobs/jobs.py
@@ -184,6 +184,10 @@ class Job(PipelineItem, CaptureInitParams):
         """Add file this job depends on."""
         self._accessed_files.add(filename.path)
 
+    @property
+    def accessed_files(self) -> set[str]:
+        return set(self._accessed_files)
+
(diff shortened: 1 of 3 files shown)


pisek/jobs/jobs.py
```diff
@@ -184,6 +184,10 @@ def _access_file(self, filename: TaskPath) -> None:
         """Add file this job depends on."""
         self._accessed_files.add(filename.path)
 
+    @property
+    def accessed_files(self) -> MutableSet[str]:
+        return self._accessed_files
```
SvizelPritula, line 189: What's the point of making this a property instead of a public field?
We're returning a `MutableSet` anyway, so it seems it's allowed to modify it from the outside.

SvizelPritula (changes requested):

SvizelPritula (approved):
LGTM

PR #271, opened 2024-12-20
Tests folder layout

### LINKED ISSUE #270: Restructure `tests/` directory

5K1PY:
Fixes #270

### DIFF
--- a/pisek/config/task_config.py
+++ b/pisek/config/task_config.py
@@ -500,12 +500,6 @@ class SolutionConfig(BaseEnv):
     def convert_auto(cls, value: str, info: ValidationInfo) -> str:
         if value == "@auto":
             value = info.data.get("name", "")
-
-        if value.startswith("_"):
-            raise PydanticCustomError(
-                "invalid_solution_source",
-                f"Solution source must not start with '_'",
-            )
         return value
 
     @field_validator("subtasks", mode="after")
--- a/pisek/config/task_config.py
+++ b/pisek/config/task_config.py
@@ -476,6 +476,11 @@ class SolutionConfig(BaseEnv):
                     "invalid_solution_name",
                     f"Solution name must not contain '{banned_char}'",
                 )
+        if value.startswith("_"):
+            raise PydanticCustomError(
+                "invalid_solution_name",
+                f"Solution name must not start with '_'",
+            )
         return value
 
     @field_validator("primary", mode="before")
@@ -494,7 +499,13 @@ class SolutionConfig(BaseEnv):
     @classmethod
     def convert_auto(cls, value: str, info: ValidationInfo) -> str:
         if value == "@auto":
-            return info.data.get("name", "")
+            value = info.data.get("name", "")
+
+        if value.startswith("_"):
+            raise PydanticCustomError(
+                "invalid_solution_source",
+                f"Solution source must not start with '_'",
+            )
         return value
 
     @field_validator("subtasks", mode="after")
--- a/pisek/config/task_config.py
+++ b/pisek/config/task_config.py
@@ -152,6 +152,11 @@ class TaskConfig(BaseEnv):
         else:
             return [name for name, sol in self.solutions.items() if sol.primary][0]
 
+    @computed_field  # type: ignore[misc]
+    @property
+    def primary_solution_file(self) -> str:
+        return self.solutions[self.primary_solution].raw_source
+
     def get_solution_by_source(self, source: str) -> Optional[str]:
         sources = (
             name for name, sol in self.solutions.items() if sol.raw_source == source
--- a/pisek/task_jobs/data/testcase_info.py
+++ b/pisek/task_jobs/data/testcase_info.py
@@ -71,8 +71,6 @@ class TestcaseInfo(yaml.YAMLObject):
             path = input_path.to_output()
 
         if solution is not None:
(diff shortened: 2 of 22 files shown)


pisek/task_jobs/data/testcase_info.py
```diff
@@ -48,11 +48,31 @@ def mixed(name: str) -> "TestcaseInfo":
     def static(name: str) -> "TestcaseInfo":
         return TestcaseInfo(name, 1, TestcaseGenerationMode.static, False)
 
-    def input_path(self, env: Env, seed: Optional[int] = None) -> TaskPath:
+    def input_path(
+        self, env: Env, seed: Optional[int] = None, solution: Optional[str] = None
```
SvizelPritula, line 52: What is the `solution` parameter for?

pisek/task_jobs/data/testcase_info.py
```diff
@@ -48,11 +48,31 @@ def mixed(name: str) -> "TestcaseInfo":
...
+    def reference_output(
+        self, env: Env, seed: Optional[int] = None, solution: Optional[str] = None
+    ) -> OutputPath:
+        is_static = self.generation_mode == TestcaseGenerationMode.static
+
+        input_path = self.input_path(env, seed, solution=env.config.primary_solution)
+        if is_static:
+            path = OutputPath.static(input_path.replace_suffix(".out").name)
+        else:
+            path = input_path.to_output()
```
SvizelPritula, line 71: What's different about static outputs?

pisek/utils/paths.py
```diff
@@ -106,74 +113,69 @@ def data_path(env: "Env", *path: str) -> "TaskPath":
...
 
-    @staticmethod
-    def invalid_file(env: "Env", name: str, seed: int) -> "TaskPath":
-        name = os.path.splitext(name)[0]
-        return TaskPath.invalid_path(env, f"{name}.{seed:x}.invalid")
 
-    @staticmethod
-    def output_path(env: "Env", *path: str) -> "TaskPath":
-        return TaskPath.data_path(env, OUTPUTS_SUBDIR, *path)
+class SanitizeablePath(TaskPath):
```
SvizelPritula, line 124: Sanitizeable is not a real word, Sanitizable is.

SvizelPritula (changes requested):

SvizelPritula:
It might be a good idea to prevent users from making a solution called `_inputs`, for example, by prohibiting solution names from statring with `_`. Although, I don't expect anyone to make such a solution.

pisek/task_jobs/data/testcase_info.py
5K1PY: Refernce static outputs are static, where others are generated by primary solution.

pisek/task_jobs/data/testcase_info.py
5K1PY: To getting input in directory corresponding to right solution

pisek/task_jobs/data/testcase_info.py
SvizelPritula: Does this function return the path for a solution-specific copy of the reference output, or to a single shared instance of the reference output?

5K1PY: Specific solution copy if `solution` is not `None`, otherwise `_inputs` instance.

SvizelPritula: I personaly found this functions logic to be hard to follow, but I think I understand it now.

pisek/config/task_config.py
```diff
@@ -494,7 +499,13 @@ def convert_yesno(cls, value: str, info: ValidationInfo) -> bool:
...
     def convert_auto(cls, value: str, info: ValidationInfo) -> str:
         if value == "@auto":
-            return info.data.get("name", "")
+            value = info.data.get("name", "")
+
+        if value.startswith("_"):
+            raise PydanticCustomError(
+                "invalid_solution_source",
+                f"Solution source must not start with '_'",
+            )
```
SvizelPritula, line 508: Is there a reason for this?

5K1PY: What in particular you are objecting?

SvizelPritula: Names cannot start with a `_` to prevent collisions with reserved folder names, but I don't see a reason for the source file name not to be able to strart with `_`.

5K1PY: We name folders according to sources, not names. As to why enforce same thing is enforced on names, is that it otherwise produces confusing error messages:
```
In section [all_solutions], key 'source' (in config file /home/skipy/dev/pisek/fixtures/sum_cms/cms-defaults):
  Solution source must not start with '_': '@auto'
  ```

SvizelPritula: True, I forgot about that.
Why are we naming the folders according to the source file, anyway?

SvizelPritula (approved):
LGTM

pisek/config/task_config.py
5K1PY: > Why are we naming the folders according to the source file, anyway?

Uups, that probably doesn't work once we allow for same source be different solutions differentiated by arguments. Gonna rewrite that.

SvizelPritula (approved):
LGTM

PR #272, opened 2025-01-01
Rename subtask to test

5K1PY:
Unifies subtask & test -> test.

I was not sure if we want to rename them too in the cms part. What do you think?


SvizelPritula (changes requested):
I don't think we need to change the CMS module, since it actually operates on subtasks.

SvizelPritula (approved):
LGTM

PR #273, opened 2025-01-07
ConfigDescription: Change checking unused keys

5K1PY:
1. Now checks keys according to `config-description` and not to their usage. This allows keys to be overshadowed by other keys.
2. Improved error messages for context-specific keys.
3. Made key recommendations more predictable. It can now recommend key that is not allowed in this context, but in such cases you probably need to look up config documentation anyway.

### DIFF
--- a/pisek/config/config_description.py
+++ b/pisek/config/config_description.py
@@ -55,7 +55,7 @@ class ConfigSectionDescription:
 
     def similarity(self, section: str) -> int:
         if self.similarity_function is None:
-            return 5
+            return 5 * (self.section != section)
         else:
             return self.similarity_function(self.section, section)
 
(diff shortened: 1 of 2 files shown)


pisek/config/config_description.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, section: str) -> None:
 
     def similarity(self, section: str) -> int:
         if self.similarity_function is None:
-            return 5
+            return 5 * (self.section != section)
```
SvizelPritula, line 58: Wouldn't this be clearer?

```suggestion
            return 5 * if self.section != section else 0
```

SvizelPritula (changes requested):

PR #274, opened 2025-01-15
Run sections

### LINKED ISSUE #134: Override limits per solution

5K1PY:
RunSections allow us to:
- Specify converte limits for each program / program category separately.
- Change run time arguments.
- Specify subdirectory for each program category (not only solutions).

And lays the ground work for BuildSections specifiyng how to compile a program.

Fixes #134


pisek/config/task_config.py
```diff
@@ -154,13 +187,12 @@ def primary_solution(self) -> str:
 
     @computed_field  # type: ignore[misc]
     @property
-    def primary_solution_file(self) -> str:
-        return self.solutions[self.primary_solution].raw_source
+    def primary_solution_run(self) -> str:
+        # TODO
```
SvizelPritula, line 191: Do what?

pisek/config/task_config.py
```diff
@@ -154,13 +187,12 @@ def primary_solution(self) -> str:
...
+        return self.solutions[self.primary_solution].run
 
-    def get_solution_by_source(self, source: str) -> Optional[str]:
-        sources = (
-            name for name, sol in self.solutions.items() if sol.raw_source == source
-        )
+    def get_solution_by_run(self, source: str) -> Optional[str]:
+        sources = (name for name, sol in self.solutions.items() if sol.run == source)
```
SvizelPritula, line 195:
```suggestion
    def get_solution_by_run(self, run: str) -> Optional[str]:
        sources = (name for name, sol in self.solutions.items() if sol.run == run)
```

pisek/config/task_config.py
```diff
@@ -228,16 +270,41 @@ def load_dict(configs: ConfigHierarchy) -> ConfigValuesDict:
...
+                    sol["run"],
+                    configs,
+                )
 
         args["limits"] = LimitsConfig.load_dict(configs)
         args["cms"] = CMSConfig.load_dict(configs)
         args["checks"] = ChecksConfig.load_dict(configs)
 
+        args["solution_time_limit"] = configs.get_from_candidates(
+            [("run_solution", "time_limit"), ("run", "time_limit")]
```
SvizelPritula, line 293: It might be a good idea to look for `run_primary_solution`, in case someone only configures `run_primary_solution` and `run_secondary_solution`

SvizelPritula (changes requested):
It looks pretty good. I haven't had the chance to try it yet.

docs/config-v3-documentation
```diff
@@ -228,40 +234,57 @@ headers=src/a-plus-b.h
...
+#                                          [run]
+#       [run_gen]          [run_checker]          [run_judge]                         [run_solution]
+#                                                                    [run_primary_solution] [run_secondary_solution]
+# [run_gen_{program}] [run_checker_{program}] [run_judge_{program}]              [run_solution_{program}]
+
+exec=@auto
+# Filename of the program to execute (without suffix)
+# @auto expands to {program} (default)
+args=
+# Additional arguments for program. Appended at the end of normally given arguments.
```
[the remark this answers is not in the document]
5K1PY: That's intentional. This way you can override arguments that came before.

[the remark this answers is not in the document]
SvizelPritula: Sure, but Pisek never passes any flags, only positional arguments. And positional arguments conventionally go after flags, which is probably what users will specify most often. The second use for the `args` parameter would be to specify a subcommand, which also always goes in front of positional arguments.

pisek/config/task_config.py
5K1PY: I'm not sure. In case of opendata tasks, you set stricter limit for primary solution, but the real time limit is much higher.

pisek/config/task_config.py
5K1PY: Can't remember. It's unused, so I'm deleting it.

docs/config-v3-documentation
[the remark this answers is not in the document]
5K1PY: Fair enough.

pisek/config/task_config.py
SvizelPritula: That wouldn't be a problem if `run_primary_solution` (or possibly `run_secondary_solution`) had a lower priority than `run_solution`, right?

5K1PY: Yeah, but which should have higher priority?

5K1PY: I think we should keep it this way, because you can choose whether limit for primary/secondary solution is the task limit. (By not setting it.)

SvizelPritula: I guess it's unclear which one should be used, so it's probably better not to guess.

5K1PY: So my idea is this.

Competitions usually have the main time limit - set in `[run_solution]`.

Then you might want to override primary or secondary.
- Primary when you have stricter requirements (opendata tasks).
- Secondary bit more lax (for checking they are not close to the limit).

In either case you consider the timelimit in `run_solution` as the competition limit and want to compare against that.

SvizelPritula (approved):
LGTM

PR #277, opened 2025-01-27
Caching rework

5K1PY:
Ran away from yaml to pickle. 🥒

Now using much better format + performance boost, because loading from yaml was really slow.


SvizelPritula (changes requested):

SvizelPritula (approved):
LGTM

PR #282, opened 2025-02-04
Docs fixing

5K1PY:
Various documentation improvements.

### DIFF
--- a/docs/config-v3-documentation
+++ b/docs/config-v3-documentation
@@ -39,7 +39,7 @@ static_subdir=static_tests/
 
 [tests]
 
-in_gen=src/gen
+in_gen=gen
 # Reference to the run section specifying how to run the generator (if empty, only static inputs are used)
 # See [run] for more
 
--- a/docs/config-v3-documentation
+++ b/docs/config-v3-documentation
@@ -103,7 +103,7 @@ judge_type=cms-batch
 judge_needs_in=0
 judge_needs_out=0
 # Only for task_type=batch and out_check=judge
-# Set to 1 if judge needs input/output (default)
+# Set to 1 if judge needs input/correct output (default)
 
 tokens_ignore_newlines=0
 # Only for out_check=tokens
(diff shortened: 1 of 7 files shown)


docs/config-v3-documentation
```diff
@@ -26,6 +26,24 @@ name=a-plus-b
 use=organization-config
 # Config to use defaults from (defaults to none)
 
+# Values of keys are loaded in the following way:
+# 1. We try to find given (section,key) in the topmost config
+# 2. If not found, we go defaults-config up.
```
SvizelPritula, line 31: What?

docs/config-v3-documentation
```diff
@@ -26,6 +26,24 @@ name=a-plus-b
 use=organization-config
 # Config to use defaults from (defaults to none)
 
+# Values of keys are loaded in the following way:
+# 1. We try to find given (section,key) in the topmost config
```
SvizelPritula: What does "topmost" mean in this section?

docs/config-v3-documentation
```diff
@@ -26,6 +26,24 @@ name=a-plus-b
 use=organization-config
 # Config to use defaults from (defaults to none)
 
+# Values of keys are loaded in the following way:
+# 1. We try to find a given (section,key) in the this config
+# 2. If not found, we go to the config specified by use key.
+# 3. If we are in the topmost config, we try searching for the defaulting
+#    section and key of this key and we reset to the task config.
+# 4. If there is no defaulting key, we either return the default value, or fail if there is none.
```
SvizelPritula, line 34: 1. "the this" doesn't make sense, it needs to be "the task" or "this".
2. It's "the use key", not just "use key".
3. The order of operations in step three is unintuitive.

```suggestion
# Values of keys are loaded in the following way:
# 1. We try to find a given (section,key) in the task config
# 2. If not found, we go to the config specified by the use key.
# 3. If we are in the topmost config, we reset to the task config an
#    try searching for the defaulting section and key of this key.
# 4. If there is no defaulting key, we either return the default value, or fail if there is none.
```

SvizelPritula (changes requested):

SvizelPritula (approved):
LGTM.

PR #286, opened 2025-02-06
Config: Move static_subdir to [tests] and name to [cms]

SvizelPritula:
As pointed out by Martin Mareš, `static_subdir` doesn't really belong in `[task]`, and `name` is only needed for the CMS importer.

### DIFF
--- a/docs/config-v3-documentation
+++ b/docs/config-v3-documentation
@@ -20,9 +20,6 @@ version=v3
 
 # You can use 'pisek update' to update the config to the highest version.
 
-name=a-plus-b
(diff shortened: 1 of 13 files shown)


5K1PY (changes requested):
This changes v2 in backwards incompatible way, no?

SvizelPritula:
> This changes v2 in backwards incompatible way, no?

Yes.

SvizelPritula:
> Well, we shouldn't do that.

I'll add it to the migration script.

5K1PY (approved):

PR #296, opened 2025-02-12
ConfigV3Docs: Document bool values

5K1PY:


docs/config-v3-documentation
```diff
@@ -116,17 +123,18 @@ judge_type=cms-batch
 judge_needs_in=0
 judge_needs_out=0
 # Only for task_type=batch and out_check=judge
-# Set to 1 if judge needs input/correct output (default)
+# Whether judge needs input/correct output (bool)
+# Both default to true value
```
SvizelPritula, line 127: I'm pretty sure "true value" would need an article. "true" is used commonly on its own.

```suggestion
# Whether the judge needs the input/correct output (bool)
# Both default to true
```

docs/config-v3-documentation
```diff
@@ -116,17 +123,18 @@ judge_type=cms-batch
...
 
 tokens_ignore_newlines=0
 # Only for out_check=tokens
-# If set to 1, newline characters will be ignored when checking the output,
+# If set to true value, newline characters will be ignored when checking the output,
```
SvizelPritula, line 131:
```suggestion
# If set to true, newline characters will be ignored when checking the output,
```

docs/config-v3-documentation
```diff
@@ -116,17 +123,18 @@ judge_type=cms-batch
...
 # as if they were any other whitespace characters
-# If set to 0, newline characters are only ignored at the end of the file (default)
+# If set to false value, newline characters are only ignored at the end of the file (default)
```
SvizelPritula, line 133:
```suggestion
# If set to false, newline characters are only ignored at the end of the file (default)
```

docs/config-v3-documentation
```diff
@@ -116,17 +123,18 @@ judge_type=cms-batch
...
 
 tokens_ignore_case=0
 # Only for out_check=tokens
-# If set to 1, ASCII characters will be compared in a case-insensitive manner (defaults to 0)
+# If set to true value, ASCII characters will be compared in a case-insensitive manner (defaults to false)
```
SvizelPritula, line 137:
```suggestion
# If set to true, ASCII characters will be compared in a case-insensitive manner (defaults to false)
```

docs/config-v3-documentation
```diff
@@ -147,7 +155,7 @@ shuffle_mode=lines
 
 shuffle_ignore_case=0
 # Only for out_check=shuffle
-# If set to 1, ASCII characters will be compared in a case-insensitive manner (defaults to 0)
+# If set to true value, ASCII characters will be compared in a case-insensitive manner (defaults to false)
```
SvizelPritula, line 158:
```suggestion
# If set to true, ASCII characters will be compared in a case-insensitive manner (defaults to false)
```

docs/config-v3-documentation
```diff
@@ -315,7 +323,7 @@ output_max_size=5
 # Please be careful with this section as disabling checks can make
 # task making into a minefield
 
-# Values are on/off: 
+# Each value is bool:
```
SvizelPritula, line 326:
```suggestion
# Each value is a bool:
```

SvizelPritula (changes requested):

SvizelPritula (approved):

PR #298, opened 2025-02-14
CMS: Fix automatic computation of title and submission format

SvizelPritula:


5K1PY (changes requested):
Let's remove `@name` please (and change it to empty string). Also it would be nicer if cms commands would complain about missing name.

SvizelPritula:
> Let's remove `@name` please (and change default to empty string).

I'm not sure that's a good default. A task with an empty title looks weird in the CWS GUI, and a task without any files to submit becomes unsubmittable.

Additionaly, I don't see the value in requiring users to always specify the file name, since it's gonna be the task name (with dashes replaced with underscores) with the language extention appended most of the time.

> Also it would be nicer if cms commands would complain about missing name.

They do.

5K1PY:
Oh right, name != title.

5K1PY (approved):

PR #301, opened 2025-02-16
Change seed length to 16 digits

5K1PY:


fixtures/sum_kasiopea/gen.cpp
```diff
@@ -17,6 +19,7 @@ int main(int argc, char* argv[]) {
     }
 
     int diff = atoi(argv[1]) - 1;
+    cerr << strlen(argv[2]) << endl;
```
SvizelPritula, line 22: Is there a reason to print the length of the seed to stderr?

fixtures/sum_kasiopea/gen.cpp
```diff
@@ -1,6 +1,8 @@
 #include <iostream>
+#include <string.h>
 #include <string>
 #include <random>
+#include <cassert>
```
SvizelPritula, line 5: Is anything declared in `cassert` used in the generator?

SvizelPritula, line 5: Why does one import of a C header use the `[name].h` format, while another uses the `c[name]` format?

SvizelPritula (changes requested):

SvizelPritula (approved):

PR #312, opened 2025-03-11
Solution: Fix overlapping inputs

5K1PY:
When inputs overlap between tests by using `in_globs` directly (and not predeccessors), it could lead to wrong cancelations and testing logs.

### DIFF
--- a/pisek/task_jobs/solution/manager.py
+++ b/pisek/task_jobs/solution/manager.py
@@ -64,6 +64,7 @@ class SolutionManager(TaskJobManager, TestcaseInfoMixin):
         )
         self._compile_job = compile_
 
+        self._sols: dict[TaskPath, RunSolution] = {}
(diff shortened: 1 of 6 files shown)


pisek/task_jobs/solution/manager.py
```diff
@@ -76,12 +77,19 @@ def _get_jobs(self) -> list[Job]:
     def _skip_testcase(
         self, testcase_info: TestcaseInfo, seed: Optional[int], test: int
     ) -> bool:
```
SvizelPritula, line 79: I'm not sure what the `_skip_testcase` method is supposed to be doing. It checks some condition, but also does other side effects?

SvizelPritula (approved):
The code LGTM, but I haven't tested it.

5K1PY:
I have tested it more than usual.

PR #318, opened 2025-03-15
Judge: Allow fuzzing to be turned off

### LINKED ISSUE #314: Add checks switch for judge fuzzing

SvizelPritula:
Fixes #314. Unfortunately, I missed Pi day.


docs/config-v3-documentation
```diff
@@ -345,6 +345,10 @@ one_input_in_each_nonsample_test=off
 # Checks that each test (excluding samples) contains exactly one input (defaults to off)
 # Useful for opendata tasks
 
+fuzz_test_judge=off
```
5K1PY, line 348: I would keep the convention that the value is equal to default.
```suggestion
fuzz_test_judge=on
```

5K1PY (changes requested):
I am unsure of the name as fuzzing is by definition testing. Personally I would call it `judge_fuzzing`.

SvizelPritula:
> I am unsure of the name as fuzzing is by definition testing. Personally I would call it `judge_fuzzing`.

I renamed it to `judge_handles_fuzzed_outputs`, since all other checks are named after a property.

5K1PY (approved):

PR #331, opened 2025-03-18
BaseEnv: Speedup

5K1PY:


SvizelPritula (approved):
I don't know how this works, so I can only really judge it stylistically. Stylistically, it looks OK.

PR #336, opened 2025-03-23
PisekV1: Fix nonexistent seed

5K1PY:


pisek/task_jobs/generator/pisek_v1.py
```diff
@@ -148,7 +148,8 @@ def _gen(self) -> None:
         )
         if result.kind != RunResultKind.OK:
             raise self._create_program_failure(
-                f"{self.generator} failed on input {self.testcase_info.name}, seed {self.seed:016x}:",
+                f"{self.generator} failed on input {self.testcase_info.name}"
+                + (":" if self.seed is None else f"seed {self.seed:016x}:"),
```
SvizelPritula, line 152: Shouldn't there be some kind of space?

```suggestion
                + (":" if self.seed is None else f", seed {self.seed:016x}:"),
```

SvizelPritula (changes requested):

SvizelPritula (approved):

PR #338, opened 2025-03-23
Util: Clean only accessed dirs

5K1PY:
This improves caching in cases when you test solution and then test the whole task.


SvizelPritula (approved):
LGTM

PR #340, opened 2025-03-30
CMS: Add flag to override time limit

SvizelPritula:
This PR allows for setting the time limit when importing with a flag:

```
pisek cms create -t 10
```

### DIFF
--- a/pisek/__main__.py
+++ b/pisek/__main__.py
@@ -81,12 +81,19 @@ def main(argv):
         )
     )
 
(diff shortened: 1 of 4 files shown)


5K1PY:
What was the reason for not using `env.timeout` for this?

SvizelPritula:
> What was the reason for not using `env.timeout` for this?

That only exists for the `test` subcommand.

5K1PY (approved):

PR #342, opened 2025-03-30
docs: improve documentation for timeout in Env class

### LINKED ISSUE #341: Fix env.timeout documentation
https://github.com/piskoviste/pisek/blob/77c7f7541f9f135346838b9685d66cdffc53adf4/pisek/env/env.py#L49

felipecastillo-b:
This PR improves the documentation for the timeout attribute in the Env class to clarify its usage and the effect of the timeout value on task testing.

Fixes #341

### DIFF
--- a/pisek/env/env.py
+++ b/pisek/env/env.py
@@ -46,7 +46,7 @@ class Env(BaseEnv):
         strict: Whether to interpret warnings as failures
         testing_log: Whether to produce testing_log.json after running
         solutions: List of all solutions to be tested
-        timeout: Timeout for task in seconds. Overrides task config if specified. (Must be >= 0)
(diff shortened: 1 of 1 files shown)


pisek/env/env.py
```diff
@@ -46,7 +46,7 @@ class Env(BaseEnv):
         strict: Whether to interpret warnings as failures
         testing_log: Whether to produce testing_log.json after running
         solutions: List of all solutions to be tested
-        timeout: Timeout for (overrides config)
+        timeout: Timeout for task in seconds. Overrides task config if specified. (Must be >= 0)
```
5K1PY, line 49:
```suggestion
        timeout: Timeout for solutions in seconds. Overrides task config if specified. (Must be >= 0)
```
There are other programs that timeout doesn't affect, so I would prefer this.

PR #346, opened 2025-04-02
CMS: Make unsubmittable solutions only a warning

SvizelPritula:


5K1PY (approved):
I don't like that `resolve_solution` can warn twice for the same solution,but fine.

PR #349, opened 2025-04-05
CMS: Verify config can be translated to CMS settings before importing

### LINKED ISSUE #344: CMS importer: Don't import checkers not supported in CMS
https://github.com/piskoviste/pisek/blob/b724cb9ff75e8b22e662eae3fa42750707a3401e/pisek/cms/dataset.py#L53

Should fail if `out_judge` is not `judge`, `diff` or `tokens` with default settings.

SvizelPritula:
Fixes #344.


5K1PY (approved):
LGTM :shipit:

PR #351, opened 2025-04-07
Cache: Swap cache file when completed

5K1PY:
Fixes EOFError after pisek is killed in the middle of cache dump.

### DIFF
--- a/pisek/jobs/cache.py
+++ b/pisek/jobs/cache.py
@@ -136,6 +136,7 @@ class Cache:
 
     def export(self) -> None:
         """Export cache into a file."""
-        with open(CACHE_CONTENT_FILE, "wb") as f:
+        with open(CACHE_CONTENT_FILE + ".tmp", "wb") as f:
             pickle.dump(self.cache, f)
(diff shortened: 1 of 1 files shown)


pisek/jobs/cache.py
```diff
@@ -136,6 +136,7 @@ def move_to_top(self, entry: CacheEntry):
 
     def export(self) -> None:
         """Export cache into a file."""
-        with open(CACHE_CONTENT_FILE, "wb") as f:
+        with open(CACHE_CONTENT_FILE + ".tmp", "wb") as f:
             pickle.dump(self.cache, f)
+        os.rename(CACHE_CONTENT_FILE + ".tmp", CACHE_CONTENT_FILE)
```
SvizelPritula, line 141: The path could be extracted into a variable:

```suggestion
        temporary_path = CACHE_CONTENT_FILE + ".tmp"
        with open(temporary_path, "wb") as f:
            pickle.dump(self.cache, f)
        os.rename(temporary_path, CACHE_CONTENT_FILE)
```

SvizelPritula (changes requested):

SvizelPritula (approved):

PR #353, opened 2025-04-09
Env: timeout -> time-limit

### LINKED ISSUE #343: Unify timeout and time-limit to time-limit
```py
    parser_test.add_argument(
        "--timeout",
        "-t",
        type=float,
        help="override time limit for solutions to TIMEOUT seconds",
    )
```
```py
        parser.add_argument(
            "--time-limit",
            "-t",
            type=float,
            help="override the time limit when importing to TIME_LIMIT seconds",
        )
```
Are called different, when they are the same thing.

(Also rename in `Env`.)

5K1PY:
Fixes #343

### DIFF
--- a/pisek/visualize.py
+++ b/pisek/visualize.py
@@ -224,7 +224,7 @@ class SolutionResults:
 
         return fails
 
-    def get_timeout_range(self, num: int) -> tuple[float, float]:
+    def get_time_limit_range(self, num: int) -> tuple[float, float]:
         results = self.get_by_test()[num]
         times = [0] + list(map(lambda r: r.time, results))
         times.sort()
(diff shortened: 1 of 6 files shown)


pisek/visualize.py
```diff
@@ -224,7 +224,7 @@ def check_all(self) -> list[str]:
 
         return fails
 
-    def get_timeout_range(self, num: int) -> tuple[float, float]:
+    def get_timelimit_range(self, num: int) -> tuple[float, float]:
```
SvizelPritula, line 227: Why is time limit spelled as one word here and only here?

SvizelPritula (changes requested):

SvizelPritula (approved):

PR #355, opened 2025-04-10
Docs: Remove mentions of removed limitations when checking the results of CMS submissions

### LINKED ISSUE #352: Update the CMS documentation
Some parts of it are no longer true:
> Note that CMS doesn't differentiate between wrong answers and timeouts or runtime errors,
so any error will be reported as "wrong".

SvizelPritula:
Fixes #352.


5K1PY (approved):
LGTM

PR #358, opened 2025-04-21
Making log level as configurable Parameter

AathiKrishna21:
Log level can be set through env.
updated .vscode/launch.json

### DIFF
--- a/pisek/utils/util.py
+++ b/pisek/utils/util.py
@@ -16,9 +16,11 @@
 
 import os
 import shutil
+import logging
 from typing import Optional
 
 from pisek.utils.paths import BUILD_DIR, TESTS_DIR, INTERNALS_DIR
+from pisek.utils.errors import ConfigError
 
 
 class ChangedCWD:
@@ -54,3 +56,19 @@ def clean_non_relevant_files(accessed_files: set[str]) -> None:
             path = os.path.join(root, file)
             if root in accessed_dirs and path not in accessed_files:
                 os.remove(path)
+
+
+def log_level_mapper(log_level: str) -> int:
+    log_level = log_level.lower()
+    if log_level == "debug":
+        return logging.DEBUG
+    elif log_level == "info":
+        return logging.INFO
+    elif log_level == "warning":
+        return logging.WARNING
+    elif log_level == "error":
+        return logging.ERROR
+    elif log_level == "critical":
+        return logging.CRITICAL
+    else:
+        ConfigError(f"Invalid log level - {log_level}!, Please set a valid log level.")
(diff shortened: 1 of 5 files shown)


pisek/utils/util.py
```diff
@@ -54,3 +56,19 @@ def clean_non_relevant_files(accessed_files: set[str]) -> None:
...
+    elif log_level == "info":
+        return logging.INFO
+    elif log_level == "warning":
+        return logging.WARNING
+    elif log_level == "error":
+        return logging.ERROR
+    elif log_level == "critical":
+        return logging.CRITICAL
+    else:
+        ConfigError(f"Invalid log level - {log_level}!, Please set a valid log level.")
```
5K1PY, line 74: You really don't want to do this as this does nothing. It would be better to use `fatal_user_error` from `utils/text.py` we added just recently:
```py
fatal_user_error(f"Invalid log level: '{log_level}'")
```

5K1PY (changes requested):
Sorry for the late reply. Good work overall, but few things need fixes.

AathiKrishna21:
Hi @5K1PY I have updated the changes based on the review comments.

5K1PY:
We use [black](https://github.com/psf/black) as our formatter. Running `black .` in top level directory should be enough to reformat the code.

Can you (before reformatting) revert the last commit (`601ccbf`) as it does edit things that there is no need to edit?

AathiKrishna21:
Updated with the changes!

pisek/utils/util.py
```diff
@@ -54,3 +56,21 @@ def clean_non_relevant_files(accessed_files: set[str]) -> None:
             path = os.path.join(root, file)
             if root in accessed_dirs and path not in accessed_files:
                 os.remove(path)
+
+
+def log_level_mapper(log_level: str) -> Union[int, None]:
```
5K1PY, line 61:
```suggestion
def log_level_mapper(log_level: str) -> int:
```
Sorry, this is a mistake on my side. Can you add typing to `fatal_user_error`?
```py
def fatal_user_error(msg, *args, **kwargs) -> NoReturn:
```

AathiKrishna21: updated

pisek/utils/text.py
```diff
@@ -15,28 +15,29 @@
...
     return text + (length - len(text)) * pad_char
 
 
-def pad_left(text: str, length: int, pad_char: str = " "):
+def pad_left(text: str, length: int, pad_char: str = " ") -> str:
     return pad(text[::-1], length, pad_char)[::-1]
 
 
-def eprint(msg, *args, **kwargs):
+def eprint(msg, *args, **kwargs) -> NoReturn:
```
5K1PY, line 35:
```suggestion
def eprint(msg, *args, **kwargs) -> None:
```
`NoReturn` means exit from program. In this case `eprint` returns None.

PR #361, opened 2025-04-21
Makefile strategy

### LINKED ISSUE #356: Support Makefile build-strategy
In this strategy `make` should be run to create executables.

5K1PY:
Fixes #356

### DIFF
--- a/pisek/task_jobs/builder/strategies.py
+++ b/pisek/task_jobs/builder/strategies.py
@@ -111,6 +111,24 @@ class BuildStrategy(ABC):
         except FileNotFoundError:
             raise PipelineItemFailure(f"Missing tool: {tool}")
 
+    def _run_subprocess(self, args: list[str], program: str, **kwargs) -> None:
+        self._check_tool(args[0])
+
+        logger.debug("Building '" + " ".join(args) + "'")
+        comp = subprocess.Popen(args, **kwargs, stderr=subprocess.PIPE)
+        while comp.stderr is not None:
+            line = comp.stderr.readline().decode()
+            if not line:
+                break
+            self._print(line, end="", stderr=True)
+
+        comp.wait()
+        if comp.returncode != 0:
+            raise PipelineItemFailure(
+                f"Build of {program} failed.\n"
+                + tab(self._env.colored(" ".join(args), "yellow"))
+            )
+
 
 class BuildScript(BuildStrategy):
     @classmethod
@@ -133,25 +151,6 @@ class BuildBinary(BuildStrategy):
     @classmethod
     def applicable_on_directory(cls, build: "BuildConfig", directory: str) -> bool:
         return False
-
-    def _run_compilation(self, args: list[str], program: str, **kwargs) -> str:
-        self._check_tool(args[0])
-
-        logger.debug("Compiling '" + " ".join(args) + "'")
-        comp = subprocess.Popen(args, **kwargs, stderr=subprocess.PIPE)
-        while comp.stderr is not None:
-            line = comp.stderr.readline().decode()
-            if not line:
-                break
-            self._print(line, end="", stderr=True)
-
-        comp.wait()
-        if comp.returncode != 0:
-            raise PipelineItemFailure(
-                f"Compilation of {program} failed.\n"
-                + tab(self._env.colored(" ".join(args), "yellow"))
-            )
-        return self.target
 
 
 class Python(BuildScript):
@@ -209,12 +208,13 @@ class C(BuildBinary):
             "-fdiagnostics-color=" + ("never" if self._env.no_colors else "always")
         )
 
-        return self._run_compilation(
+        self._run_subprocess(
             ["gcc", *self.sources, "-o", self.target, "-I."]
             + c_flags
             + self._build_section.comp_args,
             self._build_section.program_name,
         )
+        return self.target
 
 
 class Cpp(BuildBinary):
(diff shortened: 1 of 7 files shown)


pisek/task_jobs/builder/strategies.py
```diff
@@ -249,10 +250,31 @@ def applicable_on_files(cls, build: "BuildConfig", sources: list[str]) -> bool:
...
+
+    @classmethod
+    def applicable_on_directory(cls, build: "BuildConfig", directory: str) -> bool:
+        return os.path.exists(os.path.join(directory, "Makefile"))
+
+    def _build(self) -> str:
+        directory = os.listdir()[0]
+        with ChangedCWD(directory):
+            self._run_subprocess(["make"], self._build_section.program_name)
+            if not os.path.isdir("out"):
```
SvizelPritula, line 275: Why did you choose `out` as the directory name? I don't think I've ever seen a build directory named like that. Plus, I don't think the name makes much sense. `out` implies it contains outputs of the program, but it doesn't - It contains the runnable program itself. I suppose it contains the output of the compiler, but by the same logic every directory or file not created by hand could be called `out`. What about `build` or `bin`?

gollux: I would expect that `build` contains temporary files generated during building (e.g., object files), not necessarily the final output. `bin` implies executable files, but the output of the build process is not limited to executables (e.g., Python libraries usually aren't).

During building, `out` is really the output of the build process. When the built program is executed later, the directory is no longer called that.

If you have a better name, tell us please. But I don't think that either `build` or `bin` is better.

SvizelPritula: Sure, `build` does often also contain artifacts, but I think it's pretty common for it to (also) contain the build target. `bin` also makes sense - the output of any Pisek build will be one or more executables. Sure, sometimes there will be supporting files, but those can also be in the `bin` folder. It doesn't match the FSH definition of the folder, but the Dotnet build system does that.

Sure, Pisek renames and moves the directory before executing the binary, but during development, you may very well want to run build and run the command manually, in which case calling the directory `out` doesn't make much sense. I think it makes sense to pick a meaningfull name for when the program is ran outside of Pisek, since when it's built by Pisek, the name is hidden from the user anyway.

pisek/task_jobs/builder/strategies.py
```diff
@@ -249,10 +250,31 @@ def applicable_on_files(cls, build: "BuildConfig", sources: list[str]) -> bool:
...
+    def applicable_on_files(cls, build: "BuildConfig", sources: list[str]) -> bool:
+        return False
+
+    @classmethod
+    def applicable_on_directory(cls, build: "BuildConfig", directory: str) -> bool:
+        return os.path.exists(os.path.join(directory, "Makefile"))
+
+    def _build(self) -> str:
+        directory = os.listdir()[0]
+        with ChangedCWD(directory):
```
SvizelPritula, line 273: It might make sense to delete the target directory (currently `out`) first. I don't know if we want to rely on `make`s logic for determining when to rebuild a file.

gollux: The build should be executed in a fresh copy of the source tree, so `out` should not exist yet. Pisek should scream when it exists (somebody has put it in the source tree). Pisek should also create `out` before running `make`, because directory dependencies in Makefiles are an unnecessary hassle.

pisek/task_jobs/builder/strategies.py
gollux: Well, I wouldn't draw inspiration from the Dotnet build system, since Dotnet is quite alien to the UNIX world. (In fact, when I see Dotnet doing something, I immediately think if doing the _opposite_ is the right choice ;)) Besides Dotnet, I don't remember seeing any serious build system putting non-executable things into `bin`. It's not FHS, it's UNIX tradition that goes much deeper.

Anyway, while I don't consider `build` to be a good choice, it's certainly an acceptable one.

pisek/task_jobs/builder/strategies.py
SvizelPritula: > The build should be executed in a fresh copy of the source tree

It's very possible that a users working tree will contain an `out` directory. How else are you meant to debug your program or Makefile? Or do you mean that the user should clean his worktree before running Pisek? Because that would render Pisek's caching pointless.

pisek/task_jobs/builder/strategies.py
gollux: Re running things outside Pisek: I don't think that there is any difference between `out` and `build` - the user has to remember that they need to refer to a subdirectory and what is the subdirectory's name.

pisek/task_jobs/builder/strategies.py
gollux: I assumed that when users experiment with stand-alone builds, they also clean the trash (i.e., the destination directory, not the whole worktree).

Well, we might make it a bit more convenient by ignoring a pre-existing `out`, but it actually interferes with caching in a different way: the cache compares the whole contents of the source directory when deciding whether a rebuild is needed. Therefore if you leave anything in `out`, you invalidate this part of the cache. But I think it's a minor problem. (Still, I consider it good taste to clean the remnants of all experiments ;))

pisek/task_jobs/builder/strategies.py
gollux: I like `target`.

pisek/task_jobs/builder/strategies.py
SvizelPritula: This remains an issue.

Though we could put it off for later and fix it properly, say with a `.pisekignore` file, or some implicit strategy-defined exclude pattern.

SvizelPritula: We have decided to put this off to a later time.

SvizelPritula (approved):
LGTM

PR #362, opened 2025-04-21
feat(program): Change cwd of a program to be temporary directory

5K1PY:


SvizelPritula (approved):
LGTM

SvizelPritula:
It could be nice to assert that the directory is empty, as I don't imagine we want programs to write to CWD.

5K1PY:
> It could be nice to assert that the directory is empty, as I don't imagine we want programs to write to CWD.

In further plans, this directory will contain required files (like inputs/outputs) for judges, so program can't access task directory at all. But that's for later.

PR #379, opened 2025-07-06
Check that all outputs on each test are not the same

### LINKED ISSUE #371: Check that outputs aren't identical
On every subtask

5K1PY:
Fixes #371


docs/config-v3-documentation
```diff
@@ -180,6 +180,10 @@ skip_validation=false
 # If true, do not validate this test
 # Defaults to false
 
+different_outputs=true
+# checks that there not all of the primary solution's outputs are the same (if there are at least 2 testcases)
```
SvizelPritula, line 184:
```suggestion
# checks that not all of the primary solution's outputs are the same (if there are at least 2 testcases)
```

fixtures/sum_cms/gen_same_out.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+# Generates same answer for all inputs should have the same
```
SvizelPritula, line 2:
```suggestion
# Generates inputs that all have the same answer
```

docs/config-v3-documentation
SvizelPritula: Wouldn't it make sense for this key to be in the checks session? We need some convention for per-test checks.

SvizelPritula (changes requested):

docs/config-v3-documentation
```diff
@@ -185,9 +185,21 @@ predecessors=
...
 
-skip_validation=false
-# If true, do not validate this test
-# Defaults to false
+
+# Keys for enabling/disabling per test checks that the task must satisfy
+# See also [checks] section for whole task checks
+#
+# Please be careful with disabling checks as that can transform
+# task preparation into a minefield
```
SvizelPritula, line 193:
```suggestion
# Keys for enabling/disabling per-test checks that the task must satisfy
# See also the [checks] section for whole-task checks
#
# Please be careful when disabling checks as it can transform
# task preparation into a minefield
```

docs/config-v3-documentation
```diff
@@ -369,26 +382,27 @@ output_max_size=5
...
-# Sections for enabling/disabling checks that task must satisfy
-
-# Please be careful with this section as disabling checks can make
-# task making into a minefield
-
+# Section for enabling/disabling entire task checks that the task must satisfy
+# See also checks keys in [tests] for per test checks
+#
+# Please be careful with disabling checks as that can transform
+# task preparation into a minefield
```
SvizelPritula, line 389:
```suggestion
# Section for enabling/disabling whole-task checks that the task must satisfy
# See also the checks keys in [tests] for per-test checks
#
# Please be careful when disabling checks as it can transform
# task preparation into a minefield
```

SvizelPritula (approved):
LGTM. Wasn't a dot used in cache keys somewhere? Won't that cause issues?

SvizelPritula (changes requested):
Sorry, that was meant to be a change request.

5K1PY:
> Wasn't a dot used in cache keys somewhere? Won't that cause issues?

Yes, but we already changed that to use `tuple[str]` instead. Additionally, it saves field names not config keys.
https://github.com/piskoviste/pisek/blob/e625a908adc356b899c0af92fb23f6e8330e9fd4/pisek/env/base_env.py

SvizelPritula (approved):
LGMT

PR #380, opened 2025-07-06
Fix -vv formatting

5K1PY:
Fixes bug in double verbose formatting that considers inputs not printed inputs.


SvizelPritula (approved):
LGTM

PR #387, opened 2025-07-13
fix(jobs): Add job class name to signature

### LINKED ISSUE #386: Caching doesn't account for program types
Currently most of the program jobs have the same name, which disregards program type when caching.

One solution would be add job class name to signature.

5K1PY:
Fixes #386


5K1PY:
> We need to make sure we don't create job types with the same class name (in different modules).

Yup. I would like to have some tests for that. On the first run, all jobs are run. On the second, all are cached.

PR #396, opened 2025-07-20
feat(config): Implement points=unscored

### LINKED ISSUE #395: points=unscored
Checkers which give absolute points (although we don't implement any yet) can give them even on samples. But they should not be counted towards total score. (There are some usecases for them however.)

Therefore implement `points=unscored`. With this option set, solution would get 0 points **always**.

5K1PY:
Fixes #395


SvizelPritula (changes requested):
The implementation seems fine, but shouldn't this be documented in the sample configs?

PR #403, opened 2025-07-29
fix(build): Fail gracefully if run already exists

5K1PY:


pisek/task_jobs/builder/strategies.py
```diff
@@ -159,6 +159,10 @@ def _get_entrypoint(self, file_extension: str) -> str:
                     f"Entrypoint '{self._build_section.entrypoint}' not in sources."
                 )
 
+    def _check_no_run(self):
+        if "run" in self.inputs:
+            raise PipelineItemFailure("File 'run' already exists in sources or extras")
```
Tonyl314, line 164: The error message could be misleading – self.inputs is a listing of the top-level directory, not necessarily the sources or extras.

PR #420, opened 2025-08-24
refactor(pipeline-tools): Abstract separator_text into it's own method

5K1PY:


pisek/utils/terminal.py
```diff
@@ -31,5 +27,10 @@
 TARGET_LINE_WIDTH = min(terminal_width, 100)
 
 
+def separator_text(text: str):
+    dashes = TARGET_LINE_WIDTH - len(text) - 2
+    return "-" * (dashes // 2) + f" {text} " + "-" * ((dashes + 1) // 2)
```
SvizelPritula, line 32: Do we have to mix f-strings and concatenation like this?

SvizelPritula (approved):

PR #441, opened 2025-09-07
fix(task-job): Fix folder caching

5K1PY:


pisek/task_jobs/task_job.py
```diff
@@ -236,7 +237,7 @@ def _diff_files(self, file_a: TaskPath, file_b: TaskPath) -> str:
     def _globs_to_files(
         self, globs: Iterable[str], directory: TaskPath
     ) -> list[TaskPath]:
-        self._accessed_globs |= set(globs)
+        self._accessed_globs |= set(os.path.join(directory, g) for g in globs)
```
SvizelPritula, line 240: `os.path.join` doesn't work with `TaskPath`.

SvizelPritula (changes requested):

pisek/task_jobs/task_job.py
5K1PY: Oops, should have run the tests locally.

pisek/task_jobs/task_job.py
```diff
@@ -48,10 +48,11 @@ def _globs_to_files(
         self, globs: Iterable[str], directory: TaskPath
     ) -> list[TaskPath]:
         """Get files in given directory that match any glob."""
-        files: list[str] = sum(
-            (glob.glob(g, root_dir=directory.path) for g in globs),
-            start=[],
-        )
+        files_per_glob = [
+            glob.glob(g, root_dir=directory.path, recursive=True, include_hidden=True)
```
SvizelPritula, line 52: The same flags aren't passed when computing the job signature, so creating a file in a subdirectory doesn't invalidate the cache.

SvizelPritula (changes requested):

pisek/task_jobs/task_job.py
5K1PY: Fixed

SvizelPritula (approved):
LGTM

PR #447, opened 2025-09-23
Docs/mkdocs material

5K1PY:
Set up material for mkdocs documentation and added overall documentation improvements.

Blocked on #443


docs/task-overview.md
```diff
@@ -82,8 +83,7 @@ graph TD;
 Interactive task should be used when part of the input should remain hidden
 from the contestant's solution. For example when it can ask questions about the input.
```
Tonyl314, line 84:
```suggestion
An interactive task is used when a part of the input should remain hidden
from the contestant's solution. For example, when it can ask questions about the input.
```

docs/task-parts/batch-checker.md
```diff
@@ -4,6 +4,7 @@ The batch checker gets the contestants output, the input and the correct output.
 It should say whether the contestant output is correct.
 
 There are various types of checkers you can use:
+
 - `tokens` - fast, versatile file equality checker
 - `diff` - file equality checker based on the `diff` command line tool (don't use, as it has a quadratic time complexity)
 - `judge` - custom checker
```
Tonyl314, line 10:
```suggestion
- `tokens` – fast, versatile file equality checker
- `diff` – file equality checker based on the `diff` command line tool (avoid this option, as it has a quadratic time complexity)
- `judge` – custom checker
```

docs/task-parts/batch-checker.md
```diff
@@ -22,7 +23,7 @@ For the output to be correct, the tokens need to be same as in the correct outpu
 
 You can customize the tokens judge with `tokens_ignore_newlines` or `tokens_ignore_case`.
 For comparing floats, set `tokens_float_rel_error` and `tokens_float_abs_error`.
-Details can be found in [config-documentation](/docs/config-v3-documentation).
+Details can be found in [config-documentation](./config-docs).
```
Tonyl314, line 26: Broken link!
```suggestion
Details can be found in [config-documentation](../config-docs).
```

docs/task-parts/batch-checker.md
```diff
@@ -35,7 +36,9 @@ Use `shuffle_ignore_case` for case insensitivity.
 An equality checker based on the `diff` tool. Runs `diff -Bbq` under the hood.
 Ignores whitespace and empty lines.
 
-**Use is strongly discouraged because of the quadratic complexity of `diff`.**
+??? danger "This `out_check` is not recommended"
+
+    In some cases `diff` has quadratic complexity, leading to unexpectedly slow checking of outputs.
```
Tonyl314, line 41:
```suggestion
    In some cases, `diff` has quadratic time complexity, leading to unexpectedly slow checking of outputs.
```

docs/task-parts/batch-checker.md
```diff
@@ -44,6 +47,7 @@ Set `out_judge` to the path to the source code of your judge, `judge_type` to *e
 and `judge_needs_in`, `judge_needs_out` to `0`/`1` depending whether the judge needs the input and the correct output.
```
Tonyl314, line 47:
```suggestion
and `judge_needs_in`, `judge_needs_out` to `0`/`1`, depending on whether the judge needs the input and the correct output.
```

docs/task-parts/batch-checker.md
```diff
@@ -62,6 +66,14 @@ To its stderr it should write a single-line message to the contestant.
 **Unlike the CMS documentation specifies, the files should be single-line only.**
 There will be a warning otherwise.
 
+??? example "Example `cms-batch` judge"
+
+	For a [task](https://github.com/piskoviste/pisek/blob/master/examples/cms-batch) of printing *N* positive integers that sum up to *K*,
+	the judge may look like this:
+    ```py
+    --8<-- "examples/cms-batch/gen.py"
+    ```
```
Tonyl314, line 75: Wrong link!
```suggestion
	the judge may look like this:
    ```py
    --8<-- "examples/cms-batch/judge.py"
    ```
```

docs/task-parts/batch-checker.md
```diff
@@ -83,16 +95,22 @@ Otherwise, the judge should exit returncode 43.
 Optionally, the judge can write a one-line message for the contestant
 to stderr (at most 255 bytes), followed by a sequence of lines with `KEY=value` pairs.
 The following keys are allowed:
+
 - `POINTS` — Number of points awarded for this test case (used only if the exit code says "OK").
 - `LOG` — A message that should be logged.
 - `NOTE` — An internal note recorded in the database, but not visible to contestants.
```
Tonyl314, line 101: An en dash is more appropriate here.
```suggestion
- `POINTS` – Number of points awarded for this test case (used only if the exit code says "OK").
- `LOG` – A message that should be logged.
- `NOTE` – An internal note recorded in the database, but not visible to contestants.
```

docs/task-parts/batch-checker.md
```diff
@@ -83,16 +95,22 @@ Otherwise, the judge should exit returncode 43.
...
 The opendata-v1 judge is the same as opendata-v2, with the exception of using different
 returncodes, returncode 0 for a correct output and returncode 1 for a wrong output.
 
 However, usage of this judge type is discouraged as these returncodes can
 also be the result of an internal judge bug.
+
+??? danger "This `judge_type` is not recommended"
+
+    Return with exitcode 1 is very common and is for example trigger by any exception in Python.
+    This
```
Tonyl314, line 116:
```suggestion
The opendata-v1 judge is the same as opendata-v2, with the exception of using different
return codes, return code 0 for a correct output and return code 1 for a wrong output.

However, usage of this judge type is discouraged as these return codes can
also be the result of an internal judge bug.

??? danger "This `judge_type` is not recommended"

    Returning with exit code 1 is very common and is for example triggered by any exception in Python.
```

docs/task-parts/generator.md
```diff
@@ -3,18 +3,20 @@ The generator is used for generating inputs. Solutions are then run and judged o
...
 There are currently 3 generator types available:
-- pisek-v1
-- cms-old
-- opendata-v1
+
+- [pisek-v1](#pisek-v1)
+- [cms-old](#cms-old)
+- [opendata-v1](#opendata-v1)
 
 However we strongly recommend using the pisek-gen type
```
Tonyl314, line 11:
```suggestion
However, we strongly recommend using the pisek-gen type
```

docs/task-parts/generator.md
```diff
@@ -3,18 +3,20 @@ The generator is used for generating inputs. Solutions are then run and judged o
...
 
 ## Terminology
 There are two requirements for generators:
-- Generators **must** be deterministic. (For the same arguments, it should generate the same input(s).) 
-- If a generator takes a seed as an argument, the generator **should** respect the seed. (For different seeds
-  it should generate different inputs.) This can be disabled in the `[checks]` section, but be careful.
+
+- **Generators must be deterministic.** — For the same arguments, it should generate the same input(s).
+- If a generator takes a seed as an argument, the generator should **respect the seed**. — For different seeds
+  it should generate different inputs. This can be disabled in the `[checks]` section, but be careful.
```
Tonyl314, line 19:
```suggestion
- **Generators must be deterministic.** For the same arguments, it should always generate the same input(s).
- If a generator takes a seed as an argument, the generator should **respect the seed**. For different seeds,
  it should generate different inputs. This can be disabled in the `[checks]` section, but be careful.
```

docs/task-parts/generator.md
```diff
@@ -61,13 +71,21 @@ The generator must be deterministic.
...
 ```
 
 The generator should generate all input files to this directory. The generator must be deterministic.
 
+??? warning "This `gen_type` is not recommended"
+
+    We don't recommend using this generator type for these reasons:
+
+    - It's **hard to debug**, as you don't know during generation of what input the bug occurs.
+    - You need to write **your own management code** for which inputs to generate, how many times and what seeds to use.
```
Tonyl314, line 87:
```suggestion
    - It's **hard to debug**, as you don't know during the generation of what input the bug occurs.
    - You need to write **your own management code** for which inputs to generate, how many times, and what seeds to use.
```

docs/task-parts/validator.md
```diff
@@ -15,6 +16,14 @@ Where `test_num` is number of the test to which the input belongs.
 If the input is valid, the validator should exit with **returncode 42**.
 Otherwise, it should exit with any other returncode.
```
Tonyl314, line 17:
```suggestion
If the input is valid, the validator should exit with **return code 42**.
Otherwise, it should exit with any other return code.
```

docs/task-parts/validator.md
```diff
@@ -23,3 +32,8 @@ The simple-0 validator is run as follows:
 Where `test_num` is number of the test to which the input belongs.
 If the input is valid, the validator should exit with **returncode 0**.
 Otherwise, it should exit with any other returncode.
+
+??? warning "This `validator_type` is not recommended"
+
+    Return with exitcode 0 can be sometimes caused by libraries.
+    It is better to avoid this pitfall and use the simple-42 validator type instead.
```
Tonyl314, line 39:
```suggestion
Where `test_num` is the number of the test to which the input belongs.
If the input is valid, the validator should exit with **return code 0**.
Otherwise, it should exit with any other return code.

??? warning "This `validator_type` is not recommended"

    Return with exit code 0 can be sometimes caused by libraries.
    It is better to avoid this pitfall and use the simple-42 validator type instead.
```

Tonyl314 (changes requested):
There are many issues with the documentation. I checked the new additions for now, but there seem to be problems outside them as well.

docs/cheatsheet.md
```diff
@@ -0,0 +1,73 @@
...
+## Task creation
+
+Create a task skeleton:
+```bash
+pisek init
+```
+
+## Task testing
+
+Test a task:
```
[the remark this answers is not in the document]
5K1PY: It is a cheatsheet, it shouldn't be unnecessarily long. Where else would the task be?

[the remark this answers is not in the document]
Tonyl314: If we want to keep it short, it should at least be clear that the command tests a specific task (which does not need to be specified as an argument).
```suggestion
Test the task:
```

Tonyl314:
Not familiar with GitHub actions, could @SvizelPritula check those?

Tonyl314 (approved):
The docs LGTM now!

PR #448, opened 2025-09-24
Update CMS documentation

### LINKED ISSUE #425: Update CMS integration documentation
Apparently, [CMS now supports Python 3.11](https://cms.readthedocs.io/en/latest/Installation.html), therefore so complicated setup is no longer needed:

> Installing Pisek and CMS together is no easy feat.
CMS only supports Python up to version 3.9, while Pisek requires Python 3.11.
Getting CMS to work in Python 3.11 requires updating some dependencies
and making a few small changes to its source code to work around
breaking changes in Python's standard libraries.
For details, refer to the provided Dev Container Dockerfile.

SvizelPritula:
Fixes #425.


5K1PY (approved):
:shipit:

PR #450, opened 2025-09-28
Run envs

5K1PY:
Allows setting environment variables when running a program.

### DIFF
--- a/pisek/config/config_hierarchy.py
+++ b/pisek/config/config_hierarchy.py
@@ -14,7 +14,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from collections import defaultdict
 from configparser import (
     ConfigParser,
     DuplicateSectionError,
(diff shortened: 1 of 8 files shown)


pisek/config/config_hierarchy.py
```diff
@@ -41,6 +41,12 @@
 CONFIG_FILENAME = "config"
 
 
+def new_config_parser() -> ConfigParser:
+    config = ConfigParser(interpolation=None)
+    config.optionxform = lambda x: x  # type: ignore
```
SvizelPritula, line 46: This is a backwards incompatible change.

pisek/task_jobs/program.py
```diff
@@ -91,6 +91,18 @@ def __init__(self, env: Env, name: str, **kwargs) -> None:
...
 
+    @staticmethod
+    def _env_disjoint_union(
+        env1: dict[str, str], env2: dict[str, str]
+    ) -> dict[str, str]:
+        union = env1 | env2
+        if len(union) != len(env1) + len(env2):
+            duplicates = list(set(env1.keys()) & set(env2.keys()))
+            raise PipelineItemFailure(
+                "Duplicate environment variable: " + duplicates[0]
```
SvizelPritula, line 102: I'm not sure this is the best error message. Wouldn't something like "ENV_VAR is used by the protocol and cannot be overwritten" be better?

SvizelPritula (changes requested):

pisek/config/config_hierarchy.py
5K1PY: I know... Howevery it breaks thing loudly and it can be aasily fixed, so I think it's fine.

PR #457, opened 2025-09-30
fix(build): Make builds run in parallel

### LINKED ISSUE #445: Make build run parallel
Currently builds cannot run in parallel becuase they use `chdir`. Needs rewrite.

5K1PY:
Fixes #445

### DIFF
--- a/pisek/task_jobs/builder/build.py
+++ b/pisek/task_jobs/builder/build.py
@@ -14,6 +14,7 @@
 import os
 import shutil
 from typing import Optional
+import uuid
 
 from pisek.utils.text import tab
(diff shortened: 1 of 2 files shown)


SvizelPritula (approved):
LGTM

Nit: Do we actually need to restore the CWD afterwards? I'd assume that the strategy itself doesn't really do anything outside of its workdir.

5K1PY:
> Nit: Do we actually need to restore the CWD afterwards? I'd assume that the strategy itself doesn't really do anything outside of its workdir.

Cargo strategy does needed it. (Although that could be gotten rid of.) More importantly it would be a trap for our future selves.

PR #464, opened 2025-10-02
feat(main): Allow overriding name of the config file

### LINKED ISSUE #456: Add config-name override
Add `--config-name` option that overrides name of the configuration file.

5K1PY:
Fixes #456

### DIFF
--- a/pisek/env/env.py
+++ b/pisek/env/env.py
@@ -22,6 +22,7 @@ from typing import Optional
 from pisek.utils.text import eprint
 from pisek.utils.colors import ColorSettings
 from pisek.env.base_env import BaseEnv
+from pisek.config.config_hierarchy import DEFAULT_CONFIG_FILENAME
 from pisek.config.task_config import load_config, TaskConfig
 from pisek.config.select_solutions import expand_solutions, UnknownSolutions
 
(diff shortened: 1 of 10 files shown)


pisek/init.py
```diff
@@ -163,12 +177,17 @@ def from_scratch() -> int:
     return 0
 
 
-def from_template(path: str) -> int:
+def from_template(path: str, config_filename: str) -> int:
+    if config_filename in os.listdir(path):
```
SvizelPritula, line 181: Does this have any advantage over `os.path.exists`?

pisek/init.py
```diff
@@ -154,7 +165,10 @@ def from_scratch() -> int:
     config.add_section(sol_sec)
     config[sol_sec]["primary"] = "yes"
 
-    with open(CONFIG_FILENAME, "w") as f:
+    if os.path.exists(config_filename):
```
SvizelPritula, line 168: Wouldn't it be better to just pass `'x'` to open?

pisek/env/env.py
```diff
@@ -88,12 +88,13 @@ def load(
         repeat: int = 1,
         iteration: int = 0,
         pisek_dir: Optional[str] = None,
+        config_filename: str = "config",
```
SvizelPritula, line 91: Should this have a default?

SvizelPritula (changes requested):

pisek/init.py
5K1PY: Fixed in 78dcfbc

pisek/init.py
SvizelPritula: With the `x` option you can just catch the `FileExistsError`.

SvizelPritula (changes requested):
See comment.

SvizelPritula (approved):
LGTM

PR #465, opened 2025-10-02
fix(main): Respect $PISEK_DIRECTORY everywhere

5K1PY:


SvizelPritula (approved):
I feel like the PR name could use some improvement.

PR #487, opened 2025-10-14
fix(strategies): Don't fail when suffix is specified in config

### LINKED ISSUE #412: Specifing suffix in config is broken
Setting `validator=validate.py` and running ends in:
```
"Validate sample_01.in on test 0" failed:
  Program build/validate does not exist, although it should have been compiled already.
```

Also if you are reading this please please please don't use suffixes...

5K1PY:
Fixes #412


SvizelPritula (approved):
LGTM

PR #498, opened 2025-10-23
feat(config): Add config export

### LINKED ISSUE #219: Export config file
When sharing tasks it might be useful to export config file keys and values. This would get make it independent of `use`.

5K1PY:
Fixes #219

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -70,5 +70,16 @@ class TestCLIVisualize(TestCLI):
         return ["testing_log.json"]
 
 
+class TestCLIExport(TestCLI):
+    def fixture_path(self):
+        return "../fixtures/guess/"
+
+    def args(self):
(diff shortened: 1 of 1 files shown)


pisek/__main__.py
```diff
@@ -243,6 +243,13 @@ def add_argument_dataset(parser):
     config_subparsers.add_parser(
         "update", help="update config to newest version (replaces the config)"
     )
+    parser_export = config_subparsers.add_parser(
+        "export", help="creates an organization independent config"
+    )
+    parser_export.add_argument(
+        "filename",
+        help="Filename of the exported config"
```
SvizelPritula: Most other help messages start with a lowercase character.
This message would be more similar to the help message of `--config-filename`:

```suggestion
        help="name of the exported configuration file"
```

PR #551, opened 2025-12-06
Solution visuals improvments

5K1PY:


pisek/task_jobs/solution/verdicts_eval.py
```diff
@@ -42,5 +42,11 @@ def evaluate_verdicts(
             definitive &= ok
             if not ok and len(verdicts) == 1:
                 breaker = 0
+            elif ok and mode != verdict_always:
+                breaker = oks.index(True)
```
SvizelPritula: So a correct testcase breaks the solution?

pisek/task_jobs/solution/solution_manager.py
```diff
@@ -390,26 +386,34 @@ def status(
...
-            return f"({predecessor_summary}) "
-        return ""
+        if not self.previous_jobs:
+            return ""
+
+        verdicts = self._verdicts(self.previous_jobs)
+        if not verdicts:
+            return "p | "
+
+        _, _, guarantor = self._as_expected(self.previous_jobs)
```
SvizelPritula: It's kind of confusing that the final return value has a different name than inside the function.

SvizelPritula (approved):
The `check_verdicts` method is hard to understand, but otherwise ok.

PR #577, opened 2025-12-29
feat(config): Move generator_respects_seed to [tests]

### LINKED ISSUE #574: Move `generator_respects_seed` to tests sections

5K1PY:
Fixes #574


5K1PY:
> It's a bit strange that we're slowly getting rid of the [checks] section. After all, there are many other checks that could be disabled per test.

My outlook on it:
- `no_unused_inputs`, `all_inputs_in_last_test`  & `fuzzing_thoroughness` - Disabling per test doesn't have any meaning.
- `one_input_in_each_nonsample_test` - This could technically be configured per test, but doesn't make any sense to do so, as this is more of a contest system thing.
- `judge_rejects_trailing_string` - You could want this, but I can't imagine a reasonable scenario.
- `solution_for_each_test` - Disabling this one per test makes sense.

Yes, we could move the rest of the keys to `[task]` section, but I still like the current setup more.

SvizelPritula:
`fuzzing_thoroughness` could be changed per test, though it's probably not all that usefull.

Issue #575, opened 2025-12-29
Refactor `ColorSettings` singleton to not use capital `C`


5K1PY:

DAShaikh10:
Thanks @5K1PY,

Have a great New Year's Eve!

PR #579, opened 2025-12-31
Refactor `ColorSettings` Singleton to `camelCase`

DAShaikh10:
# Summary:

We change the Pascal-cased instance `ColorSettings` to snake-case `color_settings`.

# Fixes:

#575 

# Test Plan

- [x] Run the `tests` workflow as a sanity check.

image

# References:

- https://peps.python.org/pep-0008/#function-and-variable-names
- https://www.conventionalcommits.org/en/v1.0.0/

### DIFF
--- a/pisek/__main__.py
+++ b/pisek/__main__.py
@@ -33,7 +33,7 @@ from pisek.user_errors import (
 
 from pisek.utils.util import clean_task_dir, log_level_mapper
 from pisek.utils.text import eprint
-from pisek.utils.colors import colorSettings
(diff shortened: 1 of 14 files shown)


5K1PY:
Hi :wave: Looks good, I just have few remarks:
1. `snake_case` would be more pythonic instead of `camelCase` (see https://peps.python.org/pep-0008/#function-and-variable-names)
2. It would be best not to use [scope](https://www.conventionalcommits.org/en/v1.0.0/) in the commit message. Therefore the commit message would be: `refactor: ...`

PR #587, opened 2026-01-05
Decimals for time & file size

### LINKED ISSUE #582: Change limits to be Decimal instead of float

5K1PY:
Fixes #582.

The Decimals are probably not passed correctly to CMS, so needs fixes or at least double check.

### DIFF
--- a/pisek/__main__.py
+++ b/pisek/__main__.py
@@ -17,7 +17,6 @@
 
 import argcomplete
 import argparse
-from decimal import Decimal
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


SvizelPritula:
I made dataset creating convert the time limit from a decimal to a float.

pisek/cms/dataset.py
```diff
@@ -125,7 +126,9 @@ def create_dataset(
         task_type_parameters=task_params,
         score_type="GroupMin",
         score_type_parameters=score_params,
-        time_limit=time_limit if time_limit is not None else config.cms.time_limit,
+        time_limit=float(
+            time_limit if time_limit is not None else config.cms.time_limit
+        ),
```
5K1PY: We should check we are not passing zero here.

PR #618, opened 2026-01-28
fix: Remove judging dependency on time solution ran

### LINKED ISSUE #497: Sanitization & Judge is rerun when solution is rerun
Because they read the solution prerequisite which contains time the solution ran...

5K1PY:
Fixes #497


SvizelPritula (approved):
LGTM

PR #627, opened 2026-02-13
chore(version): 2.3.0

5K1PY:


SvizelPritula:
> It is technically a breaking change, but very easily fixable, (either by hand or automatically) so I don't really see the point of realeasing a major version.

Renaming old solutions and adding build sections where needed can be a lot of work. It would be nice to at least be able to to so automatically.

PR #657, opened 2026-03-21
fix(config): Do not run primary solution always in interactive tasks

5K1PY:


pisek/config/task_config.py
```diff
@@ -409,8 +409,8 @@ def load_dict(task_type: str, configs: ConfigHierarchy) -> ConfigValuesDict:
         OUT_CHECK_SPECIFIC_KEYS = [
             ((None, "judge"), "out_judge", ""),
             ((None, "judge"), "judge_type", ""),
-            ((TaskType.batch, "judge"), "judge_needs_in", "0"),
-            ((TaskType.batch, "judge"), "judge_needs_out", "1"),
+            ((TaskType.batch, "judge"), "judge_needs_in", "1"),
+            ((TaskType.batch, "judge"), "judge_needs_out", "0"),
```
SvizelPritula: What does this have to do with interactive tasks?

PR #664, opened 2026-04-13
fix(cms): Make CMS module work with typed prerequisites results

SvizelPritula:


pisek/jobs/jobs.py
```diff
@@ -383,7 +383,7 @@ def set_env(self, env: "Env") -> None:
 
     def create_jobs(self) -> list[Job]:
         """Crates this JobManager's jobs."""
-        self.result: Optional[dict[str, Any]]
+        self.result: Optional[Any]
```
5K1PY: Can we use `| None` instead of `Optional`?

5K1PY (approved):

pisek/jobs/jobs.py
SvizelPritula: Yes.

pisek/jobs/jobs.py
```diff
@@ -163,8 +162,8 @@ def _check_prerequisites(self) -> None:
 
     def add_prerequisite(
         self,
-        item: Optional["PipelineItem"],
-        name: Optional[str] = None,
+        item: "PipelineItem | None",
```
5K1PY, line 165:
```suggestion
        item: "PipelineItem" | None,
```

5K1PY:
Can we also have only one commit for this?

pisek/jobs/jobs.py
SvizelPritula:
```
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

5K1PY: okay

PR #669, opened 2026-05-01
fix(checker): Save `checker_rr` in RunOpendataJudge

5K1PY:
Fixes crashes when using opendata judge with `--stats`.


SvizelPritula (approved):
LGTM

PR #670, opened 2026-05-16
fix(config): Compute judge_needs_in/out correctly, according to task type

### LINKED ISSUE #658: Primary solution runs always on interactive tasks
Even though we don't need it to generate output.

5K1PY:
Fixes #658


SvizelPritula (approved):
LGTM

PR #677, opened 2026-06-29
feat(checker): Add codeforces-batch-v1 judge

5K1PY:


pisek/task_jobs/checker/codeforces_judge.py
```diff
@@ -0,0 +1,100 @@
...
+                self._maybe_correct_output_path(),
+                report.abspath,
+            ],
+            stderr=self.checker_log_file,
+        )
+
+        if self.checker_rr.returncode == 0:
+            return RelativeSolutionResult(
+                verdict=Verdict.ok,
+                message="OK",
```
SvizelPritula, line 80: Shouldn't the message be the content of the report file?

pisek/task_jobs/checker/checker_base.py
```diff
@@ -220,3 +223,19 @@ def _checking_message(self) -> str:
         return (
             f"output {self.output.col(self._env)} for input {self.input.col(self._env)}"
         )
+
+    @staticmethod
+    def _invalid_path(name: str) -> str:
+        return os.path.join(gettempdir(), f"the-{name}-is-not-available-{uuid4()}")
```
SvizelPritula, line 229: Nit: `testlib.h` will always open all files. While it does not seem to be a problem if a file does not exists, if it does exist, at least its size will be checked. An attacker could create a large file or directory with the right name in `/tmp/` to cause the judge to crash.

pisek/task_jobs/checker/codeforces_judge.py
5K1PY: I always treated the message as the thing shown to participants. Also report file can be multiline.

pisek/task_jobs/checker/checker_base.py
5K1PY: So you would pass all the files always?

pisek/task_jobs/checker/codeforces_judge.py
SvizelPritula: The checker comment is shown to participants, just only on samples during the contest proper.

5K1PY: And message is always shown to the participants in the opendata module.

SvizelPritula: It's not a perfect match, but I feel it's kind of like when CMS only shows you the verdict for some test cases.

Regardless of if the comment is the same thing as a message, we should record it and show it in verbose mode. (And, in the future, make it available in the opendata module.)

5K1PY: > Regardless of if the comment is the same thing as a message, we should record it and show it in verbose mode. (And, in the future, make it available in the opendata module.)

Yes, that is why this is a draft PR.

PR #678, opened 2026-07-09
fix(config): Make tests string one char shorter if has_sample_test=false

### LINKED ISSUE #667: If `has_sample_test=false`, `tests` string should contain one less characters

5K1PY:
Fixes #667

### DIFF
--- a/pisek/config/task_config.py
+++ b/pisek/config/task_config.py
@@ -186,7 +186,7 @@ class TaskConfig(BaseEnv):
         return next(sources, None)
 
     def __init__(self, **kwargs):
-        value = {"test_count": max(kwargs["test_sections"]) + 1}
+        value = {"test_numbers": list(kwargs["test_sections"].keys())}
 
         with init_context(value):
             super().__init__(**kwargs)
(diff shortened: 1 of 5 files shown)


pisek/config/task_config.py
```diff
@@ -714,6 +716,12 @@ def validate_model(self):
 
         return self
 
+    def tests(self, test: int, has_sample_test: bool) -> str:
```
SvizelPritula: This function only returns a single test, making its name confusing.

SvizelPritula: Wouldn't it be better if `SolutionSection` remembered the existence of the sample test?

pisek/config/task_config.py
```diff
@@ -186,7 +186,7 @@ def get_solution_by_run(self, run: str) -> Optional[str]:
         return next(sources, None)
 
     def __init__(self, **kwargs):
-        value = {"test_count": max(kwargs["test_sections"]) + 1}
+        value = {"test_numbers": list(kwargs["test_sections"].keys())}
```
SvizelPritula, line 189: Are you sure everything will work correctly if test numbers are not continuous?

5K1PY: They won't be, we enforce that.

pisek/config/task_config.py
SvizelPritula: It would apparently be logistically difficult.

SvizelPritula (approved):
LGTM

PR #679, opened 2026-07-18
docs(examples): Add cms-communication task

5K1PY:
Fixes #435

PR #680, opened 2026-07-18
docs(config-v3-docs): Python stubs are not supported

5K1PY:


SvizelPritula:
Doesn't #679 use them?

5K1PY:
> Doesn't #679 use them?

No, that uses library.

SvizelPritula (approved):
LGTM

